PR #4, opened 2024-08-11
Using backported support for System.Threading.Lock.

MarkCiliaVincenti:


EDITzDev:
Thanks for the contribution.
I’m aware of the new Lock type in .NET 9, I will update all the locks when it releases later this year, a backport library isn't needed.

MarkCiliaVincenti:
Meanwhile, a new version of Backport.System.Threading.Lock has come out that acts as a source generator and basically dropping the DLL as a dependency. If this new development makes you reconsider, I will amend this PR accordingly.

PR #8, opened 2025-12-01
Set default thread culture to InvariantCulture globally

strophiccat:
Force invariant culture for the server and gateway to prevent number formatting issues on non-US systems.
Ensures consistent JSON serialization and packet parsing.


src/Sanctuary.Gateway/Program.cs
```diff
@@ -14,6 +16,12 @@
 using Sanctuary.UdpLibrary.Configuration;
 using Sanctuary.UdpLibrary.Enumerations;
 
+
+CultureInfo.DefaultThreadCurrentCulture = CultureInfo.InvariantCulture;
+CultureInfo.DefaultThreadCurrentUICulture = CultureInfo.InvariantCulture;
+Thread.CurrentThread.CurrentCulture = CultureInfo.InvariantCulture;
+Thread.CurrentThread.CurrentUICulture = CultureInfo.InvariantCulture;
```
EDITzDev: These can be removed as they're redundant.

src/Sanctuary.Login/Program.cs
```diff
@@ -15,6 +17,11 @@
 using Sanctuary.UdpLibrary.Configuration;
 using Sanctuary.UdpLibrary.Enumerations;
 
+CultureInfo.DefaultThreadCurrentCulture = CultureInfo.InvariantCulture;
+CultureInfo.DefaultThreadCurrentUICulture = CultureInfo.InvariantCulture;
+Thread.CurrentThread.CurrentCulture = CultureInfo.InvariantCulture;
+Thread.CurrentThread.CurrentUICulture = CultureInfo.InvariantCulture;
```
EDITzDev: These can be removed as they're redundant.

EDITzDev (changes requested):
Thanks for the PR!

Can you also do the same for `src/Sanctuary.WebAPI/Program.cs`?

src/Sanctuary.Gateway/Program.cs
```diff
@@ -17,8 +17,6 @@
 using Sanctuary.UdpLibrary.Enumerations;
 
 
-CultureInfo.DefaultThreadCurrentCulture = CultureInfo.InvariantCulture;
-CultureInfo.DefaultThreadCurrentUICulture = CultureInfo.InvariantCulture;
 Thread.CurrentThread.CurrentCulture = CultureInfo.InvariantCulture;
 Thread.CurrentThread.CurrentUICulture = CultureInfo.InvariantCulture;
```
EDITzDev: I think you removed the wrong one here, we want to use the `CultureInfo.DefaultThreadCurrent*`.

EDITzDev (changes requested):

src/Sanctuary.Gateway/Program.cs
strophiccat: Yeah, I made a slight accident here. I'll fix it now

strophiccat: I believe the fix has been applied. Please let me know if it hasn't

PR #9, opened 2025-12-01
Implementation for Clean Shaven Style Card Fix

strophiccat:
Added a condition for empty string to be returned if modelCustomizatonId parameter is 0.


EDITzDev:
I'll have to work on this later as there are quite a few changes I need to make. You're welcome to raise an issue instead.

PR #18, opened 2026-06-19
Track friend requests in state + harden friend request validation

### LINKED ISSUE #10: Method of crashing server found: Namechange
Found by Rick, if you change your username in-game and then send a friend request to your old username, if you accept it it will cause the server to crash.

yungcomputerchair:
Few improvements related to friendships:
- Track friend requests in state
  - This fixes a vulnerability where crafted CommandPacketConfirmFriend payloads can force-create friendships without consent from the target player.
  - I added a ConcurrentSet type for this to keep it pretty.
    - Wasn't sure if it needed to be concurrent-- as it looks like all friend stuff happens on one thread-- but better safe than sorry. I also noticed a `FriendAddRequestTimedOut` message which makes it sound like we may one day need to expire them from the world ticker. I'm not comfortable enough to mess with that yet, though.
-  Prohibit sending a friend request to yourself (fixes #10)
   - The client tries to prevent you from sending the request packet out to yourself, but it only checks against name, so a name change would allow it => cause a crash when it reaches the DB.
   - Now we validate in the friend request handlers that the target player ID is not the sending player's ID. Since we track this in state, keeping the validation scoped to the request side is enough.
- Prohibit sending a friend request to someone you're already friends with
  - The client probably blocks this as well-- I didn't test it personally-- but I'm pretty sure a crafted packet could cause a similar DB crash to the name change case, since it would be a duplicate record.

### DIFF
--- a/src/Sanctuary.Core/Collections/ConcurrentSet.cs
+++ b/src/Sanctuary.Core/Collections/ConcurrentSet.cs
@@ -16,11 +16,11 @@ public class ConcurrentSet<T> : IEnumerable<T> where T : notnull
         _dictionary = new ConcurrentDictionary<T, byte>();
     }
 
-    public bool Add(T item) => _dictionary.TryAdd(item, 0);
+    public bool TryAdd(T item) => _dictionary.TryAdd(item, 0);
 
     public bool Contains(T item) => _dictionary.ContainsKey(item);
 
-    public bool Remove(T item) => _dictionary.TryRemove(item, out _);
+    public bool TryRemove(T item) => _dictionary.TryRemove(item, out _);
 
     public void Clear() => _dictionary.Clear();
 
--- a/src/Sanctuary.Core/Collections/ConcurrentSet.cs
+++ b/src/Sanctuary.Core/Collections/ConcurrentSet.cs
@@ -1,0 +1,34 @@
+using System.Collections;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+
+namespace Sanctuary.Core.Collections;
+
+public class ConcurrentSet<T> : IEnumerable<T>
+{
+    private readonly ConcurrentDictionary<T, byte> _dictionary;
+
+    /// <summary>
+    /// Initializes an instance of the ConcurrentSet class.
+    /// </summary>
+    public ConcurrentSet()
+    {
+        _dictionary = new ConcurrentDictionary<T, byte>();
+    }
+
+    public bool Add(T item) => _dictionary.TryAdd(item, 0);
+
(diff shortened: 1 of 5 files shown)


src/Sanctuary.Core/Collections/ConcurrentSet.cs
EDITzDev, line 1: The generic type `T` should have a constraint to remove the nullable warning.

You can inherit the `ConcurrentDictionary` to make the implementation simpler.
```c#
public class ConcurrentSet<T> : ConcurrentDictionary<T, byte> where T : notnull
{
    public bool TryAdd(T item) => TryAdd(item, default);
    public bool TryRemove(T item) => TryRemove(item, out _);
}
```

EDITzDev (changes requested):
Looks good, just a minor thing with the nullable warning.

src/Sanctuary.Core/Collections/ConcurrentSet.cs
yungcomputerchair: I thought of that, but the simplicity brings the internals of ConcurrentDictionary into the API surface of ConcurrentSet. It would be very confusing to someone if they tried to enumerate ConcurrentSet and got key-value entries back. I added the `notnull` constraint though.

src/Sanctuary.Core/Collections/ConcurrentSet.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    /// <summary>
+    /// Initializes an instance of the ConcurrentSet class.
+    /// </summary>
+    public ConcurrentSet()
+    {
+        _dictionary = new ConcurrentDictionary<T, byte>();
+    }
+
+    public bool Add(T item) => _dictionary.TryAdd(item, 0);
```
EDITzDev: Please rename this function to `TryAdd`, to align with the .NET Try Pattern.

src/Sanctuary.Core/Collections/ConcurrentSet.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    public bool Contains(T item) => _dictionary.ContainsKey(item);
+
+    public bool Remove(T item) => _dictionary.TryRemove(item, out _);
```
EDITzDev: Please rename this function to `TryRemove`, to align with the .NET Try Pattern.

EDITzDev (changes requested):
Sorry, two more nitpick changes.

src/Sanctuary.Core/Collections/ConcurrentSet.cs
EDITzDev: Ah yea that makes sense, thanks.

PR #21, opened 2026-06-27
Fix coin store list jumping to top after purchase

edenfps:
The ClientUpdatePacketItemAdd handler in the game client deserializes both a ClientItem and a ClientItemDefinition from the same packet payload. The server was only writing the ClientItem data, causing the client to read garbage as a malformed item definition. This triggered a global item definition update broadcast that the coin store UI listened to, causing the entire shop list to refresh and reset selection to the top.

**Change:** Added one line \clientItemDefinition.Serialize(writer);\ to serialize the full item definition into the ItemAdd packet payload, so the client receives valid definition data and the coin store does not trigger a full list refresh.


EDITzDev (changes requested):
There is another `ClientUpdatePacketItemAdd` on `PacketInGamePurchasePlaceOrderPacketHandler.cs`. Please add the ClientItemDefinition to that one too.

edenfps (commented):
Done, added to \PacketInGamePurchasePlaceOrderPacketHandler.cs\ in commit 478a19c.

PR #27, opened 2026-06-28
Add consumables system: transformations, boomboxes, food effects

Sulphural:
transformation entries with model swap, duration, cooldown. boombox items with NPC spawning and dance sequences. food effects for visual particle overlays.


src/Sanctuary.Packet/BasePlayerUpdatePacket/PlayerUpdatePacketReplaceBaseModel.cs
EDITzDev, line 1: This packet has an invalid structure, the client won't process it.

It should be;
```c#
ulong Guid
int ModelId
int CompositeEffectId
```

src/Sanctuary.Packet/BasePlayerUpdatePacket/PlayerUpdatePacketQueueAnimation.cs
EDITzDev, line 1: This packet has an invalid structure, the client won't process it.

It should be;
```c#
ulong Guid
int AnimationGroupId
int Unknown
float Unknown2
float Unknown3
float Unknown4
bool Unknown5
bool Unknown6
bool Unknown7
```

src/Sanctuary.Packet/BaseAbilityPacket/AbilityPacketExecuteClientLua.cs
EDITzDev, line 1: This packet has an invalid structure, the client won't process it.

It should be;
```c#
string Script
float Param1
float Param2
float Param3
```

src/Sanctuary.Packet.Common/ActionBarSlot.cs
EDITzDev, line 1: What's the purpose for the new `Clear` field?

src/Sanctuary.Game/Zones/IZone.cs
```diff
@@ -17,6 +17,7 @@ public interface IZone
 
     void OnClientIsReady(Player entity);
     void OnClientFinishedLoading(Player entity);
+    void RefreshPlayerCustomizations(Player player);
```
EDITzDev: What's the purpose for the new `RefreshPlayerCustomizations` function?

src/Sanctuary.Game/Zones/BaseZone.cs
EDITzDev, line 1: Please don't remove code comments, unless they implement a TODO.

src/Sanctuary.Game/Entities/Player.cs
EDITzDev, line 1: Please don't remove code comments, unless they implement a TODO.

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -52,6 +52,12 @@ public sealed class Player : ClientPcData, IEntity
 
     public int TimezoneOffset { get; set; }
 
+    public Dictionary<int, Dictionary<int, int>> ActionBarItemGuids { get; set; } = new();
+
+    public int TemporaryAppearance { get; set; }
```
EDITzDev, line 68:
```suggestion
    public int TemporaryAppearance { get; set; }
    public DateTimeOffset? TemporaryAppearanceExpiresAt { get; set; }
```

src/Sanctuary.Game/Entities/Player.cs
EDITzDev, line 223:
```suggestion
public void UpdateEveryTick()
{
        if (TemporaryAppearanceExpiresAt.HasValue &&
            TemporaryAppearanceExpiresAt.Value <= DateTimeOffset.UtcNow)
        {
            RemoveTemporaryAppearance();
        }
}
```

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -497,7 +499,6 @@ public PlayerUpdatePacketAddPc GetAddPcPacket()
             IsUnderage = Age < 18,
             IsMember = MembershipStatus != 0,
 
-            // playerUpdatePacketAddPc.TemporaryAppearance = 277;
```
EDITzDev, line 571: This needs to be reverted/updated so players that are loading a player for the first time can see the players current temporary appearance.

```suggestion
            TemporaryAppearance = TemporaryAppearance,
```

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -522,6 +523,62 @@ public PlayerUpdatePacketAddPc GetAddPcPacket()
...
+        SendToVisible(removePacket, true);
+
+        var replacePacket = new PlayerUpdatePacketReplaceBaseModel
+        {
+            Guid = Guid,
+            ModelId = Model
+        };
+        SendTunneled(replacePacket);
+    }
+
```
EDITzDev: For the Temporary Appearance to work correctly there are a few changes needed;

1. All packets sent by the Gateway need to be Tunnelled
2. The ReplaceBaseModel packet isn't needed
3. Avoid using `Task.Delay` or creating background threading tasks for timed game logic. Instead, schedule the expiration using the existing `UpdateEveryTick` or `UpdateEverySecond` update loop.

```suggestion

    public void ApplyTemporaryAppearance(int modelId, int durationMs)
    {
        TemporaryAppearance = modelId;

        if (durationMs > 0)
        {
            TemporaryAppearanceExpiresAt = DateTimeOffset.UtcNow.AddMilliseconds(durationMs);
        }

        var appearancePacket = new PlayerUpdatePacketUpdateTemporaryAppearance
        {
            Guid = Guid,
            TemporaryAppearance = modelId
        };

        SendTunneledToVisible(appearancePacket, true);
    }

    public void RemoveTemporaryAppearance()
    {
        TemporaryAppearance = 0;
        TemporaryAppearanceExpiresAt = null;

        var removePacket = new PlayerUpdatePacketRemoveTemporaryAppearance
        {
            Guid = Guid
        };

        SendTunneledToVisible(removePacket, true);
    }

```

src/Sanctuary.Game/Entities/Npc.cs
```diff
@@ -39,12 +39,8 @@ public class Npc : IEntity
 
     public float Scale { get; set; }
 
-    /// <summary>
-    /// 0 - Hostile
-    /// 1 - Neutral
-    /// 2 - Ally
-    /// </summary>
     public int Disposition { get; set; } = 1;
+    public System.Action<Player>? InteractAction { get; set; }
```
EDITzDev: This action is being assigned but isn't being invoked anywhere

EDITzDev (changes requested):
Thanks for the PR!

Submitting my review comments so far. I'm still working through `AbilityPacketClientRequestStartAbilityHandler.cs`, so I'll follow up with any additional feedback once I've finished reviewing that file.

yungcomputerchair:
Looks like there are merge conflicts; probably the NPC interaction stuff.

cillow19:
Since this is related, just some thoughts for future of the start ability handler... 

As more abilities/consumables get added, would we want to introduce an interface of some sort? The handler will get bigger and more difficult to read, and it would be nice to separate each interaction to maintain readability of base handler. Similarly, for the consumable/ability definitions, would we want to add a general parent class for abilities?

I'm not too familiar with all the abilities, or how many there are in FR. So feel free to disregard this comment or correct if team feels it's not relevant or not an appropriate approach. Just wanted to bring it up sooner rather than later.

src/Sanctuary.WebAPI/appsettings.json
```diff
@@ -9,5 +9,5 @@
     "LaunchArguments": "AssetDelivery:IndirectServerAddress=http://osfr.editz.dev/assets Portrait:UploadUrl=http://127.0.0.1:20040/image"
   },
   "AllowedHosts": "*",
-  "Urls": "http://127.0.0.1:20040"
+  "Urls": "http://0.0.0.0:5055"
```
yungcomputerchair: This is specific to your local server; we don't want this change

src/Sanctuary.WebAPI/Program.cs
```diff
@@ -19,6 +20,8 @@
 
 var builder = WebApplication.CreateBuilder(args);
 
+builder.Configuration.AddJsonFile("database.json", optional: true);
```
yungcomputerchair: Do you remember why you needed to add this?

yungcomputerchair:
> Since this is related, just some thoughts for future of the start ability handler...
> 
> As more abilities/consumables get added, would we want to introduce an interface of some sort? The handler will get bigger and more difficult to read, and it would be nice to separate each interaction to maintain readability of base handler. Similarly, for the consumable/ability definitions, would we want to add a general parent class for abilities?
> 
> I'm not too familiar with all the abilities, or how many there are in FR. So feel free to disregard this comment or correct if team feels it's not relevant or not an appropriate approach. Just wanted to bring it up sooner rather than later.

We will absolutely want this

src/Sanctuary.WebAPI/appsettings.json
Sulphural: This was accidental.

src/Sanctuary.WebAPI/Program.cs
Sulphural: Also accidental... sorry.

src/Sanctuary.Gateway/Handlers/BaseAbilityPacket/AbilityPacketClientRequestStartAbilityHandler.cs
```diff
@@ -27,16 +55,610 @@ public static bool HandlePacket(GatewayConnection connection, ReadOnlySpan<byte>
...
+        _logger.LogInformation("AbilityPacket: Id={Id} Slot={Slot}", packet.Data.Id, packet.Data.Slot);
+
+        if (packet.Data.Id == 2)
+            return HandleItemAbility(connection, packet);
+
+        connection.SendTunneled(new AbilityPacketFailed { StringId = 3079 });
+        return true;
+    }
+
+    private static bool HandleItemAbility(GatewayConnection connection, AbilityPacketClientRequestStartAbility packet)
```
yungcomputerchair, line 71: This function is HUGE. I would for sure make separate, per-category helper functions for boomboxes, cakes, etc instead of handling all of them in the top-level handler

src/Sanctuary.WebAPI/appsettings.json
yungcomputerchair: No sweat, just need them out by the time we get the PR merged.

src/Sanctuary.Database/DatabaseContext.cs
yungcomputerchair, line 1: I think we want all the changes in this file gone, too

src/Sanctuary.Gateway/Handlers/BaseAbilityPacket/AbilityPacketClientRequestStartAbilityHandler.cs
```diff
@@ -27,16 +55,610 @@ public static bool HandlePacket(GatewayConnection connection, ReadOnlySpan<byte>
...
+        {
+            connection.Player.SendTunneledToVisible(new AbilityPacketExecuteClientLua { Script = string.Empty, Param1 = 0, Param2 = 0, Param3 = 0 }, true);
+        }
+
+        if (effectId != 0)
+        {
+            var effectPacket = new PlayerUpdatePacketPlayCompositeEffect { Guid = connection.Player.Guid, CompositeEffectId = effectId, Clear = true };
+
+            if (effectDelayMs > 0)
+                Task.Delay(effectDelayMs).ContinueWith(_ => connection.Player.SendTunneledToVisible(effectPacket, true));
```
yungcomputerchair: Per @EDITzDev comment:
> Avoid using Task.Delay or creating background threading tasks for timed game logic. Instead, schedule the expiration using the existing UpdateEveryTick or UpdateEverySecond update loop.

yungcomputerchair (changes requested):
Getting there, but still needs some cleanup

yungcomputerchair (approved):
This looks way better. I'd say it's good enough to merge for now.

Since it's a pretty sizable change, I'll hold off until @EDITzDev confirms all his feedback was addressed and that it looks good to him.

StimpyDev:
Why is `Consumables.jsonc` is called .jsonc?

yungcomputerchair:
> Why is `Consumables.jsonc` is called .jsonc?

.jsonc = "**json** with **c**omments"

StimpyDev:
> > Why is `Consumables.jsonc` is called .jsonc?
> 
> .jsonc = "**json** with **c**omments"

No need for .jsonc then, let’s just use a plain .json with a "Name" or "Comment" field like we do elsewhere `Npcs.json` etc.. instead of // comments

StimpyDev:
Thanks, this is way cleaner

src/Sanctuary.Gateway/Handlers/BaseAbilityPacket/AbilityPacketClientRequestStartAbilityHandler.cs
```diff
@@ -138,6 +144,126 @@ private static bool HandleCake(GatewayConnection connection, int slot, ClientIte
...
+    private const int SillyStringAnimationId = 3351;
+    private const float SillyStringGestureSeconds = 1.5f;
+
+    // Tag-based loop, not the effect's own 5s default lifetime - free to run longer.
+    private const float SillyStringEffectSeconds = 20f;
+
+    private const float SillyStringRange = 12f;
+    private const float SillyStringRangeSquared = SillyStringRange * SillyStringRange;
+    private const int SillyStringIdleAnimId = 1;
+    private const int SillyStringCooldownMs = 3000;
```
yungcomputerchair: why are these hardcoded? they should be in JSON somewhere

src/Sanctuary.Gateway/Handlers/BaseAbilityPacket/AbilityPacketClientRequestStartAbilityHandler.cs
```diff
@@ -138,6 +144,126 @@ private static bool HandleCake(GatewayConnection connection, int slot, ClientIte
...
+            Guid = target.Guid,
+            TagId = tagId,
+            CompositeEffectId = effectId,
+            SourceGuid = player.Guid,
+        };
+
+        foreach (var recipient in recipients)
+            recipient.SendTunneled(beam);
+
+        _logger.LogInformation("Silly String: {who} sprayed {target}.", player.Name, target.Name);
```
yungcomputerchair: info isn't the right logging level for this, either remove or make trace level

src/Sanctuary.Gateway/Handlers/BaseAbilityPacket/AbilityPacketClientRequestStartAbilityHandler.cs
```diff
@@ -138,6 +144,126 @@ private static bool HandleCake(GatewayConnection connection, int slot, ClientIte
...
+
+        _ = Task.Run(async () =>
```
yungcomputerchair: Please remember not to spawn async tasks and instead track timed events in game state and handle them during tick

src/Sanctuary.Gateway/Handlers/BaseAbilityPacket/AbilityPacketClientRequestStartAbilityHandler.cs
```diff
@@ -222,6 +352,9 @@ private static bool ConsumeItem(GatewayConnection connection, ClientItem clientI
             connection.Player.Items.Remove(clientItem);
             connection.SendTunneled(new ClientUpdatePacketItemDelete { ItemGuid = clientItem.Id });
 
+            // Otherwise the still-ticking cooldown re-sends this slot as non-empty a second later.
+            connection.Player.CancelActionBarCooldown(2, actionBarSlot);
```
yungcomputerchair: This seems pretty jank; I think we need to confirm against the client that this is the correct way to handle this. I have a feeling it's not

src/Sanctuary.Packet.Common/BaseItemDefinition.cs
```diff
@@ -4,6 +4,10 @@ namespace Sanctuary.Packet.Common;
 
 public class BaseItemDefinition : ISerializableType
 {
+    // Server-side only, never on the wire - ClientItemDefinitions.json's own "Comment" field (the
+    // item's real internal name, e.g. "Blue Silly String Can"), used to identify items by keyword.
+    public string Comment { get; set; } = "";
+
     public int Id { get; set; }
```
yungcomputerchair, line 7: You shouldn't need this; the comment field is only for people reading the JSON. It doesn't need to be deserialized and the deserializer shouldn't have a problem skipping over it

yungcomputerchair (changes requested):

src/Sanctuary.Gateway/Handlers/BaseAbilityPacket/AbilityPacketClientRequestStartAbilityHandler.cs
cillow19: agreed - is there a reason there isn't another definition called something like PartyAbilities? this would be helpful since we can then build off of the definition when we want to add more party items. this is also in line with your other consumables.

src/Sanctuary.Gateway/Handlers/BaseAbilityPacket/AbilityPacketClientRequestStartAbilityHandler.cs
```diff
@@ -27,16 +61,709 @@ public static bool HandlePacket(GatewayConnection connection, ReadOnlySpan<byte>
...
+        var bestFallback2 = SillyStringRangeSquared;
+
+        foreach (var candidate in zone.Players)
+        {
+            if (candidate.Guid == player.Guid)
+                continue;
+
+            var dx = candidate.Position.X - player.Position.X;
+            var dz = candidate.Position.Z - player.Position.Z;
+            var d2 = dx * dx + dz * dz;
```
cillow19: not a big deal but is there a reason we are not using Vector4 distance calculation? since position is already vector4 type. also it would be nice to pull this into separate function in case other abilities need to calculate the nearest player too.

src/Sanctuary.Game/Consumables/SillyStringAbilities.cs
```diff
@@ -0,0 +1,36 @@
+namespace Sanctuary.Game.Consumables;
```
cillow19: Is there a reason we need this folder? i think we can put silly strings with the rest of your consumable definitions under resources

cillow19 (commented):
I just wanted to add my two cents in here since i actually got some work started on silly strings as well and was hoping to add on top of your PR once it got merged. This works too though!

i know i mentioned this before, but i am having growing concerns about the length of `AbilityPacketClientRequestStartAbilityHandler`. I am seeing a lot of common patterns in the ability sub-handlers (handleBoombox, handleFoodEffects, etc.) that could be pulled out. And I suggest that a good next step would be to create an interface for abilities + get some inherited definitions going. That way, it will be easier for other devs to build on top of this gold mine of stuff!!!

yungcomputerchair:
@cillow19's feedback here is really valuable. We can keep testing this publicly, but I think we need to further clean this up before it's ready for the main branch.

If you guys could work together to unify the ability system, that would be amazing.

src/Sanctuary.Gateway/Handlers/BaseAbilityPacket/AbilityPacketClientRequestStartAbilityHandler.cs
Sulphural: This is tested to work.

yungcomputerchair: There's a difference between working and being correct. Every packet affects the internal state of the client, and some cause side effects that aren't desired. We saw an example of this with the initial collections PR: the packet worked to show the updated mushroom count, but broke a bunch of internal character state because it wasn't the exact packet meant for that situation.

It's suspicious to me that we'd have to explicitly send action bar cancels to show the slot as non-empty; it feels like a workaround. This strikes me as the kind of thing the client is supposed to handle on its own. Either we're updating the client cooldown when we shouldn't be (which the comment here seems to suggest) or we're using the wrong mechanism entirely somewhere.

Sulphural:
> I just wanted to add my two cents in here since i actually got some work started on silly strings as well and was hoping to add on top of your PR once it got merged. This works too though!
> 
> i know i mentioned this before, but i am having growing concerns about the length of `AbilityPacketClientRequestStartAbilityHandler`. I am seeing a lot of common patterns in the ability sub-handlers (handleBoombox, handleFoodEffects, etc.) that could be pulled out. And I suggest that a good next step would be to create an interface for abilities + get some inherited definitions going. That way, it will be easier for other devs to build on top of this gold mine of stuff!!!

I have something in the works for this.. If you would like to discuss further.

EDITzDev (commented):
Let me know when you're done updating the PR so I don't waste my time reviewing it again.

Sulphural:
> Let me know when you're done updating the PR so I don't waste my time reviewing it again.

Updates are done.. Sorry.

src/Resources/CoinStoreItems.json
yungcomputerchair, line 1: This file is reporting ~78 THOUSAND line changes. I'm guessing your editor formatted it. Can you please fix it so the diff only includes what you changed in this PR?

src/Sanctuary.Gateway/Handlers/Abilities/ConsumableAbility.cs
```diff
@@ -0,0 +1,233 @@
...
+
+public sealed record AbilityServices(
+    ILogger Logger,
+    IResourceManager ResourceManager,
+    IDbContextFactory<DatabaseContext> DbContextFactory);
+
+public abstract class ConsumableAbility(AbilityServices services)
+{
+    // Static so cooldowns and effect tag ids are shared across every ability, not per instance.
+    private static readonly ConcurrentDictionary<ulong, ConcurrentDictionary<int, DateTimeOffset>> _itemCooldowns = new();
```
CarterW24: These cooldowns are global static state keyed by player guid, and entries are never removed — anyone who logs off stays in the dict forever. Since it's per-player state, could it live on Player instead (same place the recast gate ended up in #121)? That fixes the leak and drops the guid lookup. _lastSillyStringTarget in SillyStringAbility.cs has the same issue.

src/Sanctuary.Gateway/Handlers/BaseAbilityPacket/AbilityPacketClientRequestStartAbilityHandler.cs
```diff
@@ -1,8 +1,13 @@
-﻿using System;
+using System;
```
CarterW24: nit: line 1 just drops the BOM — same whitespace nit I got on #121

src/Sanctuary.Packet/BasePlayerUpdatePacket/PlayerUpdatePacketAddEffectTagCompositeEffect.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace Sanctuary.Packet;
+
+// op35 sub41 "AddEffectTagCompositeEffect" — attach a LOOPING composite effect to a character as part
+// of a status-effect tag (the visible half of a buff/debuff). GROUND TRUTH (2014-04-01 capture idx
+// 37215, the heart pickup): when the player grabbed the health powerup the server attached
+// composite 15921 (PFX_magic-heal_red_head_shower_lg_loop_raised = the looping heal shower over the
+// head + trail) to the player, keyed by tag id, sourced from the heart's guid.
+//   wire (36B): [short 35][short 41][ulong Guid][int TagId][int CompositeEffectId][ulong SourceGuid]
+//               [int 0][int 0]
+// Pair with PlayerUpdatePacketRemoveEffectTagCompositeEffect (sub42) to stop the loop.
```
CarterW24: same note ycc gave me on this exact file back in #74 — "good research notes to keep locally but aren't relevant to these serializers." These headers can go. (RemoveEffectTagCompositeEffect.cs has a smaller one too).

CarterW24: that was 5-12, I accidentally only left the comment on 12

src/Sanctuary.Gateway/Handlers/Abilities/ConsumableAbility.cs
yungcomputerchair: agreed, this is per-player state so it should live in the player class.

src/Sanctuary.Gateway/Handlers/Abilities/AbilityTargeting.cs
```diff
@@ -0,0 +1,31 @@
+using Sanctuary.Game.Entities;
+using Sanctuary.Game.Zones;
+
+namespace Sanctuary.Gateway.Handlers.Abilities;
+
+public static class AbilityTargeting
+{
+    // Nearest other player to "from" within range, excluding excludeGuid if given.
+    public static Player? FindNearestPlayer(IZone zone, Player from, float range, ulong excludeGuid = 0)
```
yungcomputerchair: Can we have this be a member function of `Player` instead of an awkward helper function? That would let you get rid of both the `zone` and `from` parameters, and then you can totally delete this file. It's currently the odd one out (only file in the ability handlers that is not an actual handler).

yungcomputerchair (changes requested):
Just one gripe and then I think this is good. I'm also a little unsure about the `Abilities` folder breaking the organization of the `Handlers` module; that folder is all packet handlers specifically. Ideally it would go in `Helpers` instead and be `Player`-oriented instead of `Connection`-oriented; then I think this would be great.

yungcomputerchair (approved):
Got a few more nits but they can be handled on `main`. Thanks for addressing all the feedback!

PR #28, opened 2026-06-28
Fix regression where "Now Entering"/welcome audio can play twice (#23)

StimpyDev:
This fixes a regression from a previous pull request.

### DIFF
--- a/src/Sanctuary.Game/Entities/IEntity.cs
+++ b/src/Sanctuary.Game/Entities/IEntity.cs
@@ -32,7 +32,7 @@ public interface IEntity : IEquatable<IEntity>, IDisposable
 
     #region Update
 
-    void UpdatePosition(Vector4 position, Quaternion rotation);
+    void UpdatePosition(Vector4 position, Quaternion rotation, bool updateZoneArea = true);
     void TeleportToZone(IZone zone, Vector4 position, Quaternion rotation);
 
     void UpdateEveryTick();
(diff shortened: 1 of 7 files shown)


StimpyDev:
Although i dont like the change in the ```IEntity Class``` because It changes the IEntity.UpdatePosition for all entities, what i can do is make a ```SyncTeleportPosition(Vector4 position, Quaternion rotation)``` function instead. but lmk.

StimpyDev:
Can i have your opinion on this? @yungcomputerchair

yungcomputerchair:
Personally, I think this is a good solution. It's how I would have done it.

yungcomputerchair (approved):

yungcomputerchair:
If you are confident that not updating the zone area is the correct fix, I'm good to merge it.

StimpyDev:
> If you are confident that not updating the zone area is the correct fix, I'm good to merge it.

Because the server is syncing with the player's position now when teleporting, before the server wouldn't known or still think the player is on that position before they teleported somewhere else, but that caused a regression with the entering welcome messages playing twice. I'm fine with the change and it fixes the issue.

yungcomputerchair:
One more question: if I teleport into, say, Wugachug with this change, will the client still show the "Now Entering Wugachug" popup?

StimpyDev:
> One more question: if I teleport into, say, Wugachug with this change, will the client still show the "Now Entering Wugachug" popup?

Yep, or if u walked into the new zone, u can test it for urself if u want

PR #29, opened 2026-06-28
Fix style card inventory sync

StimpyDev:
This change is more a matter of opinion do you want to use style cards indefinitely or not?


EDITzDev:
Updating the local player data is a fix but I'm not sure how the original game worked in terms of consuming the item when applied. We might have to ask the researchers to confirm.

StimpyDev:
You indeed lose the style card after use.

yungcomputerchair (approved):

PR #30, opened 2026-07-03
Fix zone boundary checks and login error handling

StimpyDev:
This fixes a issue with boundary checks if an player/npc went to far in the northen/eastern side of the map player's/npc's were treated as invalid tiles so they appeared invisible.

### DIFF
--- a/src/Sanctuary.Gateway/Handlers/BaseInventoryPacket/InventoryPacketEquipByGuidHandler.cs
+++ b/src/Sanctuary.Gateway/Handlers/BaseInventoryPacket/InventoryPacketEquipByGuidHandler.cs
@@ -143,7 +143,7 @@ public static class InventoryPacketEquipByGuidHandler
 
         playerUpdatePacketEquipItemChange.Attachment = clientUpdatePacketEquipItem.Attachment;
 
-        playerUpdatePacketEquipItemChange.ProfileId = connection.Player.ActiveProfileId;
+        playerUpdatePacketEquipItemChange.ProfileId = packet.ProfileId;
 
         if (!_resourceManager.ItemClasses.TryGetValue(clientItemDefinition.Class, out var itemClass))
         {
--- a/src/Sanctuary.Gateway/Handlers/BaseInventoryPacket/InventoryPacketEquipByGuidHandler.cs
+++ b/src/Sanctuary.Gateway/Handlers/BaseInventoryPacket/InventoryPacketEquipByGuidHandler.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Linq;
 
 using Microsoft.EntityFrameworkCore;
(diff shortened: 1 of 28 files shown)


src/Sanctuary.Login/Handlers/CharacterCreateRequestHandler.cs
```diff
@@ -198,7 +198,7 @@ public static bool HandlePacket(LoginConnection connection, Span<byte> data)
         {
             _logger.LogInformation("Failed to create character, unknown user.");
 
-            characterCreateReply.Result = 4;
+            characterCreateReply.Result = 3;
```
yungcomputerchair, line 201: it would be great if we had these declared as constants since there are multiple meaningful result codes

src/Sanctuary.Gateway/Handlers/BaseInventoryPacket/InventoryPacketEquipByGuidHandler.cs
```diff
@@ -143,7 +143,7 @@ public static bool HandlePacket(GatewayConnection connection, ReadOnlySpan<byte>
 
         playerUpdatePacketEquipItemChange.Attachment = clientUpdatePacketEquipItem.Attachment;
 
-        playerUpdatePacketEquipItemChange.ProfileId = connection.Player.ActiveProfileId;
+        playerUpdatePacketEquipItemChange.ProfileId = packet.ProfileId;
```
yungcomputerchair: why are these two different? using the packet value is risky because the packet can be spoofed. the active profile ID should be the source of truth; do you know why it's wrong?

yungcomputerchair (changes requested):

src/Sanctuary.Gateway/Handlers/BaseInventoryPacket/InventoryPacketEquipByGuidHandler.cs
yungcomputerchair: oh, I see now; this is the *receiving* player's active profile ID. would still prefer if we fetched the requesting player's ActiveProfileId instead of using the value from the packet.

src/Sanctuary.Login/Handlers/CharacterCreateRequestHandler.cs
StimpyDev: Look in CharacterCreateReply 

```    // 1 - Success
    // 2 - FailureUnableToCreateCharacter
    // 3 - FailureUnableToProcessRequest
    // 4 - FailureNameAlreadyTaken
    // 5 - FailureTemporaryNameAlreadyTaken
    // 6 - FailureTooManyCharactersOnAccount
    // 7 - FailureInvalidName
    // X - UnknownError```

yungcomputerchair:
I still don't understand what is causing the original bug here. Using `ActiveProfileId` for the connected player is correct for all of these. We should figure out what the real cause of the desync is. Copilot thinks the client can implicitly change the active profile ID for another character, which would leave the server with a stale one if true.

Does the client ever automatically change your profile if you, e.g., equip certain items? This is something I don't have enough knowledge of.

StimpyDev:
Change the title because i've reverted all `packet.ProfileId` to `ActiveProfileId` should be good to go now.

yungcomputerchair (approved):

yungcomputerchair:
> Change the title because i've reverted all `packet.ProfileId` to `ActiveProfileId` should be good to go now.

Thanks, if you can open an issue for the profile bug you were seeing so we don't lose it, that would be great :)

PR #32, opened 2026-07-05
Guild and guild chat functionality.

strophiccat:
Addition of the guilds. Included a change where the invite to guild option does not appear when you're not in a guild or ranked as an officer or higher. Revert if needed.


strophiccat:
sent this by accident. still needs some touch ups

PR #34, opened 2026-07-06
Fix possible player name overflow in DB + crash on character deletion due to FK constraints

yungcomputerchair:
Two distinct fixes:
- Max player name length = max firstname + space + max lastname = 16 + 1 + 16 = **33**, not 32, so player names can currently overflow. Fixed the max length and added a migration for it.
- Currently, server crashes if you delete a character that is friends with other characters.
  - The Friend/Ignore records are not deleted on cascade due to the way the tables are setup, but that means there would be records with stale foreign keys, so the delete operation fails.
  - Since the delete op is not surrounded with try/catch, this throws and crashes the server.
  - Fix is two-fold: manually delete the friend/ignore records first, and then surround with try/catch so the server doesn't go down under a failure.
    - Bonus: an early return was missing, so under failure a success result was stored. Fixed.

### DIFF
--- a/src/Sanctuary.Database.MySql/Configuration/DbCharacterConfiguration.cs
+++ b/src/Sanctuary.Database.MySql/Configuration/DbCharacterConfiguration.cs
@@ -16,7 +16,7 @@ public sealed class DbCharacterConfiguration : IEntityTypeConfiguration<DbCharac
 
         builder.Property(c => c.FirstName).IsRequired().HasMaxLength(16);
         builder.Property(c => c.LastName).IsRequired(false).HasMaxLength(16);
-        builder.Property(c => c.FullName).HasMaxLength(32).HasComputedColumnSql($"CONCAT_WS(' ', `{nameof(DbCharacter.FirstName)}`, NULLIF(`{nameof(DbCharacter.LastName)}`, ''))", true);
+        builder.Property(c => c.FullName).HasMaxLength(33).HasComputedColumnSql($"CONCAT_WS(' ', `{nameof(DbCharacter.FirstName)}`, NULLIF(`{nameof(DbCharacter.LastName)}`, ''))", true);
 
         builder.Property(c => c.Model).IsRequired();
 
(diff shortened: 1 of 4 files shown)


StimpyDev:
The missing return is already fixed in [https://github.com/Open-Source-Free-Realms/Sanctuary/pull/30](url) Should not be a problem when merging, but worth mentioning :)

StimpyDev:
Also nice fix but small suggestion for extra safety

SaveChanges() can return 0 without throwing, so the client could still get status 1 even when nothing was deleted. The try/catch handles exceptions

```csharp
if (dbContext.SaveChanges() <= 0)
{
    _logger.LogError("Failed to delete character {CharacterId}.", character.Id);
    characterDeleteReply.Status = 2;
    connection.Send(characterDeleteReply);
    return true;
}
```

yungcomputerchair:
> Also nice fix but small suggestion for extra safety
> 
> 
> 
> SaveChanges() can return 0 without throwing, so the client could still get status 1 even when nothing was deleted. The try/catch handles exceptions
> 
> 
> 
> ```csharp
> 
> if (dbContext.SaveChanges() <= 0)
> 
> {
> 
>     _logger.LogError("Failed to delete character {CharacterId}.", character.Id);
> 
>     characterDeleteReply.Status = 2;
> 
>     connection.Send(characterDeleteReply);
> 
>     return true;
> 
> }
> 
> ```

Thanks for the catch; I'll update the PR.

StimpyDev:
> > Also nice fix but small suggestion for extra safety
> > SaveChanges() can return 0 without throwing, so the client could still get status 1 even when nothing was deleted. The try/catch handles exceptions
> > ```cs
> > if (dbContext.SaveChanges() <= 0)
> > 
> > {
> > 
> >     _logger.LogError("Failed to delete character {CharacterId}.", character.Id);
> > 
> >     characterDeleteReply.Status = 2;
> > 
> >     connection.Send(characterDeleteReply);
> > 
> >     return true;
> > 
> > }
> > ```
> 
> Thanks for the catch; I'll update the PR.

No problem, be sure to take a look at my pull requests too so we can help each other out while editz is gone for the time being.

StimpyDev:
Looks good.

PR #36, opened 2026-07-06
Fix 100% CPU usage for login + gateway servers (main busyloop)

yungcomputerchair:
Main loop in both the login and gateway servers is not sleeping, so both cause 100% CPU usage. Easy fix is to just give the thread a chance to sleep if there's nothing to process.


StimpyDev:
Can be simplified by using ```if (!(_loginServer.GiveTime() | _gatewayServer.GiveTime()))
                Thread.Sleep(1);``` instead.

yungcomputerchair:
I think having the bool decomposed per-line is a lot easier to read at a glance and conveys the intention of the check more clearly.

yungcomputerchair:
this has been battle-tested on the public server; merging

PR #38, opened 2026-07-07
Replace hardcoded safe teleport with nearest warpstone

StimpyDev:
Safe teleport no longer sends players to hardcoded sanctuary coordinates. The handler picks the nearest zone POI (NotificationType == 7) from PointOfInterests.json based on the player’s X/Z position, then teleports to that POI’s spawnpostion.

Make sure u merge this one first before [merging this one](https://github.com/Open-Source-Free-Realms/Sanctuary/pull/28)


StimpyDev:
But why does it bring other commits with it?

PR #39, opened 2026-07-07
Replace hardcoded safe teleport with nearest warpstone

StimpyDev:
Safe teleport no longer sends players to hardcoded sanctuary coordinates. The handler picks the nearest zone POI (NotificationType == 7) from PointOfInterests.json based on the player’s X/Z position, then teleports to that POI’s spawnpostion.

Make sure u merge this one first before [merging this one](https://github.com/Open-Source-Free-Realms/Sanctuary/pull/28)

### DIFF
--- a/src/Sanctuary.Gateway/Handlers/PacketZoneSafeTeleportRequestHandler.cs
+++ b/src/Sanctuary.Gateway/Handlers/PacketZoneSafeTeleportRequestHandler.cs
@@ -1,10 +1,13 @@
 using System;
+using System.Linq;
 using System.Numerics;
 
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 
+using Sanctuary.Game;
 using Sanctuary.Packet;
+using Sanctuary.Packet.Common;
 using Sanctuary.Packet.Common.Attributes;
 
 namespace Sanctuary.Gateway.Handlers;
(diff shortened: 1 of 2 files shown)


src/Sanctuary.Gateway/Handlers/PacketZoneSafeTeleportRequestHandler.cs
```diff
@@ -47,4 +64,33 @@ public static bool HandlePacket(GatewayConnection connection, Span<byte> data)
 
         return true;
     }
+
+    private static PointOfInterestDefinition? FindNearestSafePointOfInterest(Vector4 playerPosition)
+    {
+        var hubPointsOfInterest = _resourceManager.PointOfInterests.Values
+            .Where(x => x.NotificationType == 7)
```
yungcomputerchair: what is NotificationType 7? would be nice to have a named constant

yungcomputerchair (approved):

src/Sanctuary.Gateway/Handlers/PacketZoneSafeTeleportRequestHandler.cs
StimpyDev: Those are the major zone hubs in PointOfInterests.json. Without that filter you'd teleport to the nearest cooking table or dungeon. A named constant would help readability but the logic itself is fine.

yungcomputerchair: I'm good to get this in but it would be awesome if you could switch it to a named constant. Magic numbers that are preexisting are one thing, but we should try to stay away from them in general going forward

StimpyDev: We can always do that, but for now it's fine, it doesnt change the logic.

yungcomputerchair: I know the logic is fine, but it's good practice. It's a quick change and we won't have to come back and fix it if you just put it in with the addition.

StimpyDev: Well, go ahead. I cant fix it rn, i'm not at home anymore

yungcomputerchair: There's no rush. I'll merge whenever you get to it.

StimpyDev: Or just add a comment? Either is fine tbf

yungcomputerchair (approved):
Much better, thanks!

StimpyDev:
> Much better, thanks!

😊

PR #40, opened 2026-07-07
Fix ActiveProfile crash on invalid profile ID and fix double sending guid 

StimpyDev:
ActiveProfile used ```Profiles.Single(...)```. If the ```ActiveProfileId``` in the database does not match a loaded profile (due to corrupted data, a deleted profile, or a resource mismatch), this throws an ```InvalidOperationException```often occurring during login, friend-related operations, or profile switching.

### DIFF
--- a/src/Sanctuary.Game/Entities/Player.cs
+++ b/src/Sanctuary.Game/Entities/Player.cs
@@ -36,7 +36,8 @@ public sealed class Player : ClientPcData, IEntity
     public int ChatBubbleBackgroundColor { get; set; }
     public int ChatBubbleSize { get; set; }
 
-    public ClientPcProfile ActiveProfile => Profiles.Single(x => x.Id == ActiveProfileId);
+    public ClientPcProfile ActiveProfile =>
+        Profiles.FirstOrDefault(x => x.Id == ActiveProfileId) ?? Profiles.First();
 
     public Mount? Mount { get; set; }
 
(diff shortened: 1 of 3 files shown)


src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -36,7 +36,8 @@ public sealed class Player : ClientPcData, IEntity
     public int ChatBubbleBackgroundColor { get; set; }
     public int ChatBubbleSize { get; set; }
 
-    public ClientPcProfile ActiveProfile => Profiles.Single(x => x.Id == ActiveProfileId);
+    public ClientPcProfile ActiveProfile =>
+        Profiles.FirstOrDefault(x => x.Id == ActiveProfileId) ?? Profiles.First();
```
yungcomputerchair, line 40: Can't `Profiles.FIrst()` return null if `Profiles` is empty?

yungcomputerchair: nvm, it throws actually, so the type is correct. But is that acceptable?

StimpyDev: Yes. A logged in player always has at least one profile, empty profiles mean corrupted data, so throwing an exception is fine. This primarily fixes an invalid ActiveProfileId without causing normal gameplay to crash.

PR #42, opened 2026-07-07
Spawn in static list of community-contributed NPCs

yungcomputerchair:
I compiled the spreadsheet of community-contributed NPCs into a JSON file and made it load as a resource, then spawned those NPCs into the main zone on init.

For performance, I also made the tick loop skip NPCs marked as static (which, for now, is all of them).


GroaxyVRC:
Looks good but might have to look through the NPC JSON for invalid NPCs that we accidentially logged, just so we don't have random things placed around the world

yungcomputerchair:
> Looks good but might have to look through the NPC JSON for invalid NPCs that we accidentially logged, just so we don't have random things placed around the world

What should I look for? Bad model IDs?

GroaxyVRC:
It's hard to tell but have the NPC's spawned in and walk around the map in search of these placed models, an example being the boombox placed in the water at seaside

yungcomputerchair:
> It's hard to tell but have the NPC's spawned in and walk around the map in search of these placed models, an example being the boombox placed in the water at seaside

I don't know the game well enough to judge what is correctly placed and what is misplaced.. these are the placements currently in use on the public server. Can we merge for now and refine these over time with more community feedback?

GroaxyVRC:
Yeah, it should be fine to merge and log which npcs are out of place and need removing, and update the json overtime

PR #43, opened 2026-07-08
Fix Fire Hawk Mount Stats nothing major

GroaxyVRC:
Fix Fire Hawk Mount Stats

### DIFF
--- a/src/Resources/Mounts.json
+++ b/src/Resources/Mounts.json
@@ -370,7 +370,15 @@
     "ImageSetId": 8556,
     "ModelId": 4590,
     "Stats": {
-      "MaxMovementSpeed": 12.5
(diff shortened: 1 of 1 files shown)


yungcomputerchair (approved):
Curious where you got these numbers from?

StimpyDev:
Also groaxy, there is a bug where snowhill music is missing in snowhill and i think in shrouded glade too? can you find that bug and push that in here too? Thanks

GroaxyVRC:
The fix for the snowhill music and other areas are to be done on Editz asset server

StimpyDev:
> The fix for the snowhill music and other areas are to be done on Editz asset server

Roger that groaxy ;)

PR #44, opened 2026-07-08
Admin Controls in Game Chat

cillow19:
## Admin Controls 
I have added game chat commands for moderation options. Respective options can only be seen and used by mods/admins. 
Admins can: 
- Promote users by name to moderators, demote users by name from moderators
- Ban, Mute, Unban, Unmute
- Bans and mutes can have time limits to them. (e.g. !admin ban <player name> [# of Minutes])
Moderators can:
- Ban, Mute, Unban, Unmute
- Admins can ban/mute moderators, but moderators cannot ban/mute admins

The command prefix is "!admin". A mod/admin can use "!admin help" for a list of possible commands according to their permissions. 

### Codebase Changes 
- Migrations to User table with moderation columns
- ChatCommandRegistry.cs - holds commands and implementations along with edge cases
- Updates to PacketChatHandler.cs - mute check 
- Updates to PacketLoginHandler.cs - ban check
- Add some admin logging (if needed)

Appreciate any kind of feedback and suggestions. Hope this helps!


src/Sanctuary.Database.Entities/DbUser.cs
```diff
@@ -18,7 +18,10 @@ public sealed class DbUser
     public bool IsLocked { get; set; }
     public bool IsMember { get; set; }
     public bool IsAdmin { get; set; }
-
+    public bool IsMod { get; set; }
+    public DateTimeOffset? LockedUntil { get; set; }
+    public bool IsMuted { get; set; }
+    public DateTimeOffset? MutedUntil { get; set; }
```
yungcomputerchair: I think we should get rid of `IsMuted` and just use the optional `MutedUntil` as the mute signal. If `MutedUntil` is null, they're not muted. Otherwise, they're muted if the time is in the future. This prevents any desync between two vars and keeps it a single source of truth.

src/Sanctuary.Gateway/Handlers/PacketChat/ChatCommandRegistry.cs
```diff
@@ -0,0 +1,388 @@
...
+    Player = 0,
+    Mod = 1,
+    Admin = 2
+}
+
+public delegate void ChatCommandHandler(GatewayConnection connection, string[] args);
+
+public sealed record ChatCommandDefinition(ChatCommandRole RequiredRole, string Usage, ChatCommandHandler Handler);
+
+public static class ChatCommandRegistry
```
yungcomputerchair, line 27: I really like how you set this registry up. I think in the future, when we add more commands, we will probably want to group and route into separate files (i.e. have an `AdminChatCommand.cs` with all the functions and then add them to the registry from there), but since these are the only commands for now I think this is good.

src/Sanctuary.Gateway/Handlers/PacketChat/ChatCommandRegistry.cs
```diff
@@ -0,0 +1,388 @@
...
+    private static bool TryParseTarget(string[] args, out string parsedTargetName, out DateTimeOffset? parsedUntilValue, out string? error)
+    {
+        parsedTargetName = string.Empty;
+        parsedUntilValue = null;
+        error = null;
+
+        if (args.Length == 0)
+            return false;
+
+        if (args.Length > 1 && int.TryParse(args[^1], out var minutes))
```
yungcomputerchair, line 76: Huh, TIL. I didn't even know about `^` in C#. Thanks for teaching me something new!

src/Sanctuary.Database.Entities/DbUser.cs
yungcomputerchair: This would also eliminate things like `TryClearExpiredMute` because you wouldn't have to go out of your way to keep the two in-sync. You can just leave an expired mute in the DB and it wouldn't cause any harm.

src/Sanctuary.Gateway/Handlers/PacketChat/PacketChatHandler.cs
```diff
@@ -25,6 +30,53 @@ public static void ConfigureServices(IServiceProvider serviceProvider)
...
+
+        ChatCommandRegistry.Initialize(_zoneManager, _dbContextFactory, adminLogger);
+    }
+
+    private static bool TryClearExpiredMute(GatewayConnection connection)
+    {
+        Player player = connection.Player;
+
+        // mute is not expired if the current time has not exceeded mute time or if player is permanently muted (MutedUntil is null)
+        if (player.MutedUntil == null || player.MutedUntil > DateTimeOffset.UtcNow)
```
yungcomputerchair: I think this check is inverted, should be:
```suggestion
        if (player.MutedUntil == null || player.MutedUntil < DateTimeOffset.UtcNow)
```

yungcomputerchair (changes requested):
This looks mostly good! I think you can clean up a lot by making `IsMuted` a method on player backed by `MutedUntil` instead of a separate signal in the DB.

yungcomputerchair:
I also just noticed there's a conflict in `Player.cs`. It looks like an easy one to resolve, but let me know if you need help with it.

src/Sanctuary.Database.Entities/DbUser.cs
cillow19: Got it. This makes sense. I also realize now that permanently muting someone does not make sense lol. I have removed IsMuted and made it a method in Player.cs instead. I have also removed TryClearExpiredMute.

src/Sanctuary.Gateway/Handlers/PacketChat/ChatCommandRegistry.cs
cillow19: Agreed, something to keep in mind later down the line!

src/Sanctuary.Gateway/Handlers/PacketChat/PacketChatHandler.cs
cillow19: Thanks for suggestion, don't think this is needed anymore since I removed TryClearExpiredMute - but feel free to verify IsMuted in Player.cs!

cillow19:
> I also just noticed there's a conflict in `Player.cs`. It looks like an easy one to resolve, but let me know if you need help with it.

Conflicts have been resolved - feel free to verify! Thanks for the review!! Appreciate it!!

StimpyDev:
Make sure you check when building the solution works fine without errors. 

Also doing /admin ban is way easier then doing !admin ban

yungcomputerchair (changes requested):
You applied the feedback perfectly, thank you! Sorry to bother; I just have one more request while we're here.

Now that you've added `LockedUntil`, we have the same antipattern with `IsLocked` as we did with `MutedUntil`/`IsMuted` that you fixed.

Do you think you could similarly replace `IsLocked` entirely? The migration would basically look like "if IsLocked is true, lock them until the end of time".

cillow19:
> Make sure you check when building the solution works fine without errors.
> 
> Also doing /admin ban is way easier then doing !admin ban

I did not want to mix together player emote commands/prefixes and moderation commands. But that's just my personal design choice. 

Thanks for feedback and yeah, I'll make sure the solution builds!

yungcomputerchair:
> I did not want to mix together player emote commands/prefixes and moderation commands. But that's just my personal design choice. 

I agree with your design choice here. Custom commands should have a distinct prefix from the client's commands.

yungcomputerchair (approved):
Looks good to me. The `IsLocked` value in the DB is trashed, but that's fine as we don't have any locked players in the public DB currently. We need to clean up our DB story soon anyway.

yungcomputerchair:
Thanks for the PR and welcome aboard!

cillow19:
Yay! Thanks for the thorough review! I appreciate it a lot :DDD

PR #46, opened 2026-07-08
Additions to Guild branch

strophiccat:
adding the missing guild support and fixes the main guild bugs around invites, renames, chat, and login loading.
  guild renames now validate names, block duplicates, and only notify members in that guild. invites now check
  permissions, guild size, and whether the player is already in a guild. guild chat is scoped to the guild and respects
  ignores. it also cleans up stale guild member rows so old/broken db records don’t crash players when they log in.


src/Sanctuary.Core/Collections/ConcurrentSet.cs
strophiccat, line 1: purpose of this is to provide a simple thread-safe collection for tracking unique values. i.e. guild/player state without allowing duplicate entries and without the risk of HashSet<T> breaking under concurrent access.

src/Sanctuary.Gateway/Handlers/BaseNameChangePacket/ChangeNameRequestPacketHandler.cs
```diff
@@ -99,45 +103,4 @@ private static ChangeNameResponse OnChangeCharacterName(GatewayConnection connec
 
         return ChangeNameResponse.Pending;
     }
-
-    private static ChangeNameResponse OnChangeGuildName(GatewayConnection connection, ChangeNameRequestPacket packet)
```
strophiccat, line 103: removed old rename path because renames now use dedicated guild rename handlers. old function placed too much trust in incoming packets.

src/Sanctuary.Packet/BaseGuildPacket.cs
```diff
@@ -21,12 +21,12 @@ public virtual void Write(PacketWriter writer)
 
     public virtual bool TryRead(ref PacketReader reader)
     {
-        if (!reader.TryRead(out short opCode) && opCode != OpCode)
```
strophiccat, line 24: logical error.
should be or, not and.. each character only needs to be one allowed type and cannot be all of them at once. same for below

src/Sanctuary.Gateway/Handlers/BaseNameChangePacket/CheckNamePacketHandler.cs
```diff
@@ -93,46 +92,11 @@ private static CheckNameResponse OnCheckCharacterName(GatewayConnection connecti
...
-
         var taken = dbContext.Characters.Any(x => x.FirstName == packet.Name.FirstName && x.LastName == packet.Name.LastName);
 
         if (taken)
             return CheckNameResponse.Taken;
 
         return CheckNameResponse.Available;
     }
-
-    private static CheckNameResponse OnCheckGuildName(GatewayConnection connection, CheckNamePacket packet)
```
strophiccat, line 105: removed because guild names are validated through dedicated check handler. old function was treating a guild name like a character name with first and last name, which isn't how guild names are checked

src/Sanctuary.Gateway/Handlers/BaseNameChangePacket/CheckNamePacketHandler.cs
```diff
@@ -93,46 +92,11 @@ private static CheckNameResponse OnCheckCharacterName(GatewayConnection connecti
...
-    {
-        // TODO: Implement the following checks (https://archive.ph/3DB0L)
-        //  3 - Profane
-        // 11 - IllegalCharacters
```
strophiccat, line 109: i think these checks should be configurable per server. considering different servers may have varying policies on profanities and character rules.

StimpyDev: The PR i did for this, includes a NameFilter.txt if it's empty there are no checks on profane names. And i defaulted it to empty because for this reason, people can add whatever names they wanna block.

src/Sanctuary.Core/Collections/ConcurrentSet.cs
yungcomputerchair: This is the same one from the main branch, right?

yungcomputerchair:
Any reason you changed a bunch of the doc comments to normal inline comments? There are also some variable name changes that make the diff a little noisy; maybe a merging artifact?

src/Sanctuary.Core/Collections/ConcurrentSet.cs
strophiccat: yes, same purpose. it seems needed for guild state updates to behave properly under concurrent player/gateway events, at least from what i saw while testing.

strophiccat:
> Any reason you changed a bunch of the doc comments to normal inline comments? There are also some variable name changes that make the diff a little noisy; maybe a merging artifact?

the reason for this is personal preference. revert changes if necessary. the variable renames were just cleanup/readability while merging. in retrospect the diff is nosier than ideal. no behaviour changes were intended

yungcomputerchair:
> > Any reason you changed a bunch of the doc comments to normal inline comments? There are also some variable name changes that make the diff a little noisy; maybe a merging artifact?
> 
> the reason for this is personal preference. revert changes if necessary. the variable renames were just cleanup/readability while merging. in retrospect the diff is nosier than ideal. no behaviour changes were intended

sounds good, thank you very much for the PR. I'll clean it up a little and get it merged.

PR #47, opened 2026-07-08
ignore functionality improvements

strophiccat:
ignore fixes on top of current main. ignore and unignore now work against saved character records instead of only online players, meaning players can manage ignores even when another character is offline. system feedback for failed ignore action is clearer now as well. 

additionally ignoring yourself or someone already present on your friends list is blocked. ignore records are removed when a character is deleted and old orphaned ignore records are cleaned up so that stale database rows do not crash players while loading.

### DIFF
--- a/src/Sanctuary.Login/Handlers/CharacterDeleteRequestHandler.cs
+++ b/src/Sanctuary.Login/Handlers/CharacterDeleteRequestHandler.cs
@@ -63,8 +63,13 @@ public static class CharacterDeleteRequestHandler
         try
         {
             // Delete references first to avoid foreign key constraint violations.
-            var friends = dbContext.Friends.Where(f => f.FriendCharacterId == character.Id);
-            var ignores = dbContext.Ignores.Where(i => i.IgnoreCharacterId == character.Id);
+            var characterId = character.Id;
+
+            var friends = dbContext.Friends.Where(f =>
+                f.CharacterId == characterId || f.FriendCharacterId == characterId);
+
+            var ignores = dbContext.Ignores.Where(i =>
+                i.CharacterId == characterId || i.IgnoreCharacterId == characterId);
 
             dbContext.Friends.RemoveRange(friends);
             dbContext.Ignores.RemoveRange(ignores);
--- a/src/Sanctuary.Gateway/Handlers/BaseCommandPacket/CommandPacketIgnoreRequestHandler.cs
+++ b/src/Sanctuary.Gateway/Handlers/BaseCommandPacket/CommandPacketIgnoreRequestHandler.cs
@@ -9,10 +9,10 @@ using Sanctuary.Core.Helpers;
 using Sanctuary.Database;
 using Sanctuary.Database.Entities;
 using Sanctuary.Game;
+using Sanctuary.Gateway.Helpers;
 using Sanctuary.Packet;
 using Sanctuary.Packet.Common;
 using Sanctuary.Packet.Common.Attributes;
-using Sanctuary.Packet.Common.Chat;
 
 namespace Sanctuary.Gateway.Handlers;
 
@@ -48,7 +48,7 @@ public static class CommandPacketIgnoreRequestHandler
 
         if (dbCharacterToIgnore is null)
         {
-            SendSystemMessage(connection, "Player not found.");
+            ChatHelper.SendSystemMessage(connection, "Player not found.");
             return true;
         }
 
@@ -58,7 +58,7 @@ public static class CommandPacketIgnoreRequestHandler
 
         if (requesterCharacterId == targetCharacterId)
         {
-            SendSystemMessage(connection, "You cannot ignore yourself.");
+            ChatHelper.SendSystemMessage(connection, "You cannot ignore yourself.");
             return true;
         }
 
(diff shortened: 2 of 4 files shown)


src/Sanctuary.Login/Handlers/CharacterDeleteRequestHandler.cs
```diff
@@ -60,6 +60,14 @@ public static bool HandlePacket(LoginConnection connection, Span<byte> data)
...
         }
 
+        var characterId = character.Id;
+        var ignoreLinks = dbContext.Ignores
+            .Where(x => x.CharacterId == characterId || x.IgnoreCharacterId == characterId)
+            .ToList();
+
+        if (ignoreLinks.Count > 0)
+            dbContext.Ignores.RemoveRange(ignoreLinks);
+
```
yungcomputerchair: I think this will conflict with the change introduced in #34

yungcomputerchair (changes requested):
Looks good but I think it will conflict with the character deletion code on main. You can either rebase onto main + force push or just remove the character deletion part.

StimpyDev:
About the ignore improvements. These changes aren't part of the original game, right? In my opinion i think that we should keep it 1:1 to the original game.

yungcomputerchair:
> About the ignore improvements. These changes aren't part of the original game, right? In my opinion i think that we should keep it 1:1 to the original game.

Which improvements specifically? I agree that we should keep it in-line with the OG behavior but nothing stood out to me.

StimpyDev:
> > About the ignore improvements. These changes aren't part of the original game, right? In my opinion i think that we should keep it 1:1 to the original game.
> 
> Which improvements specifically? I agree that we should keep it in-line with the OG behavior but nothing stood out to me.

No need to send a system message, just a return is fine. I dont mind it but make it a public class instead, so we can use it for other things.

strophiccat:
hi, combined both solutions. i messed up the indentation but shouldn't be too difficult of a fix.

yungcomputerchair:
> hi, combined both solutions. i messed up the indentation but shouldn't be too difficult of a fix.

Looks good, I would just make SendSystemMessage a public helper (if you want to keep it) as Stimpy suggested, then this is good to go.

StimpyDev:
> > hi, combined both solutions. i messed up the indentation but shouldn't be too difficult of a fix.
> 
> Looks good, I would just make SendSystemMessage a public helper (if you want to keep it) as Stimpy suggested, then this is good to go.

I should wait before merging, like i said we dont know if this is normal behavior in the og game. We should ask the researchers. If they are fine with it with this change, then it should be good to go if u make this a public class :)

Also having the string static is not a good idea, because the game uses different locals, which u can choose in the launcher. It should match/update the local the user choose.

yungcomputerchair:
> > > hi, combined both solutions. i messed up the indentation but shouldn't be too difficult of a fix.
> > 
> > 
> > Looks good, I would just make SendSystemMessage a public helper (if you want to keep it) as Stimpy suggested, then this is good to go.
> 
> I should wait before merging, like i said we dont know if this is normal behavior in the og game. We should ask the researchers. If they are fine with it with this change, then it should be good to go if u make this a public class :)
> 
> Also having the string static is not a good idea, because the game uses different locals, which u can choose in the launcher. It should match/update the local the user choose.

The problem is that the ignore remove packet currently only takes a GUID-- no result code-- so there's no way for the client to know the reason why it happened or dispatch its own UI with the reason.

We also aren't setup for any localization; all of the locale related stuff on the server is hardcoded to `en_us`/United States. I would say that English is good for now, and we should set up a proper locale system later.

yungcomputerchair (approved):

StimpyDev:
> > > > hi, combined both solutions. i messed up the indentation but shouldn't be too difficult of a fix.
> > > 
> > > 
> > > Looks good, I would just make SendSystemMessage a public helper (if you want to keep it) as Stimpy suggested, then this is good to go.
> > 
> > 
> > I should wait before merging, like i said we dont know if this is normal behavior in the og game. We should ask the researchers. If they are fine with it with this change, then it should be good to go if u make this a public class :)
> > Also having the string static is not a good idea, because the game uses different locals, which u can choose in the launcher. It should match/update the local the user choose.
> 
> The problem is that the ignore remove packet currently only takes a GUID-- no result code-- so there's no way for the client to know the reason why it happened or dispatch its own UI with the reason.
> 
> We also aren't setup for any localization; all of the locale related stuff on the server is hardcoded to `en_us`/United States. I would say that English is good for now, and we should set up a proper locale system later.

All good :D, i agree with your statement here but let's not forget this in the future ^_^

yungcomputerchair:
> > > > > hi, combined both solutions. i messed up the indentation but shouldn't be too difficult of a fix.
> > > > 
> > > > 
> > > > Looks good, I would just make SendSystemMessage a public helper (if you want to keep it) as Stimpy suggested, then this is good to go.
> > > 
> > > 
> > > I should wait before merging, like i said we dont know if this is normal behavior in the og game. We should ask the researchers. If they are fine with it with this change, then it should be good to go if u make this a public class :)
> > > Also having the string static is not a good idea, because the game uses different locals, which u can choose in the launcher. It should match/update the local the user choose.
> > 
> > 
> > The problem is that the ignore remove packet currently only takes a GUID-- no result code-- so there's no way for the client to know the reason why it happened or dispatch its own UI with the reason.
> > We also aren't setup for any localization; all of the locale related stuff on the server is hardcoded to `en_us`/United States. I would say that English is good for now, and we should set up a proper locale system later.
> 
> All good :D, i agree with your statement here but let's not forget this in the future ^_^

I'll make an issue for it :)

PR #50, opened 2026-07-08
Limit pending outgoing friend requests to 1

StimpyDev:
I noticed u could spam friend requests so here ya go.

Also @yungcomputerchair I noticed a glitch in your server, when u add a player and u want to remove the player after, it doesnt want to remove the player u wanna remove. But for the other player it shows that i deleted them, so if the other person send a friend requests to me i get duplicated names in my friend list. I dont think it's a bug because i tried to reproduce it on a fresh server and it worked fine.

### DIFF
--- a/src/Sanctuary.Game/Entities/Player.cs
+++ b/src/Sanctuary.Game/Entities/Player.cs
@@ -45,6 +45,9 @@ public sealed class Player : ClientPcData, IEntity
     public List<IgnoreData> Ignores { get; set; } = [];
 
     public ConcurrentSet<ulong> IncomingFriendRequests { get; } = [];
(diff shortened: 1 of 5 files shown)


yungcomputerchair:
I think this is an over-engineered solution. And I don't think we need to track both incoming AND outgoing; that's two trackers for one mechanism.

Pretty sure we can achieve the same result of preventing spam by just early-returning based on the return value of `IncomingFriendRequests.TryAdd()`. 

As for the bug on the public server; what commit were you on where you couldn't repro? The public server is a few commits behind right now, so if you can't repro on main we can consider it a non-issue until we find a repro.

StimpyDev:
Bro? What did github desktop do with my commit.

PR #51, opened 2026-07-08
Send visible NPC notifications

amuralle:
## Summary
- add typed `PlayerUpdatePacketAddNotifications` and `NotificationInfo` serialization
- expose optional `Npc.Notification` metadata
- send NPC cursor relevance and batched notifications when NPCs become visible

## Why
`PlayerUpdatePacketAddNotifications` is already listed in the packet reader map, and `Player.OnAddVisibleNpcs` had dormant handling for NPC relevance/notifications. This wires that intended path so NPC-backed interactables can provide notification metadata to the client.

## Testing
- `dotnet build .\src\Sanctuary.sln --no-incremental -p:BaseOutputPath=.buildcheck\`

### DIFF
--- a/src/Sanctuary.Game/Entities/Npc.cs
+++ b/src/Sanctuary.Game/Entities/Npc.cs
@@ -61,7 +61,7 @@ public class Npc : IEntity
 
     public byte CursorId { get; set; }
 
-    // public NotificationInfo? Notification { get; set; }
+    public NotificationInfo? Notification { get; set; }
 
     public List<CharacterAttachmentData> Attachments { get; set; } = [];
 
--- a/src/Sanctuary.Game/Entities/Player.cs
+++ b/src/Sanctuary.Game/Entities/Player.cs
@@ -270,7 +270,7 @@ public sealed class Player : ClientPcData, IEntity
             SendTunneled(playerUpdatePacketAddNpc);
         }
 
-        /* var playerUpdatePacketNpcRelevance = new PlayerUpdatePacketNpcRelevance();
+        var playerUpdatePacketNpcRelevance = new PlayerUpdatePacketNpcRelevance();
 
         foreach (var npc in npcs)
         {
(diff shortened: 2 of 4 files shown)


amuralle:
FYI this is small foundational stuff for collections- which are represented as NPCs with interactables  - just dipping my toes in with this PR and making sure my styling etc is ok.

yungcomputerchair (approved):
This code looks totally fine to me, and I like how you made the batching efficient. Since this code was dormant before, I would just ask that you do some runtime testing and make sure that this doesn't open up any big holes or crashes.

yungcomputerchair:
Thanks for the PR btw!

amuralle:
can do

amuralle:
Runtime-tested this locally against the PR branch.

What I verified:
- Built the branch successfully with `dotnet build src/Sanctuary.sln`.
- Added a local-only runtime probe, not committed to the PR, that spawned two visible NPCs with `NotificationInfo` attached: one compact notification payload and one full notification payload.
- Loaded into the client and confirmed both NPCs rendered, which exercised `Player.OnAddVisibleNpcs`, `PlayerUpdatePacketNpcRelevance`, and the new batched `PlayerUpdatePacketAddNotifications` send path.
- Walked out of range and back in to force visible NPC remove/add churn, then confirmed the NPCs popped out/in without a client disconnect.
- Checked Gateway/Login/WebAPI stderr logs: no crashes or exceptions.
- Gateway continued receiving normal client metrics after the visibility churn, so the session remained healthy.

The only log noise observed was the existing class of unhandled startup/movement packets we already see during normal local play; nothing new pointed at this notification batching path.

adding SS of logs in a sec

amuralle:
e1

logs

e2


empty err.logs also

yes I use dracula theme no I will accept no teasing!

yungcomputerchair:
I use Dracula too :laughing:

PR #52, opened 2026-07-08
Prevent friend request spam

StimpyDev:


src/Sanctuary.Game/Interactions/AddFriendInteraction.cs
```diff
@@ -31,7 +31,11 @@ public void OnInteract(Player player, IEntity other)
         if (otherPlayer.Friends.Any(x => x.Guid == player.Guid))
             return;
 
-        otherPlayer.IncomingFriendRequests.TryAdd(player.Guid);
+        if (player.Friends.Any(x => x.Guid == otherPlayer.Guid))
```
yungcomputerchair: this new check is redundant; you don't need to check both directions, just one

yungcomputerchair (changes requested):

src/Sanctuary.Game/Interactions/AddFriendInteraction.cs
StimpyDev: Overlooked, my bad

PR #53, opened 2026-07-08
Update with public server resources + fix trailing comma in Mounts.json

yungcomputerchair:
Two things:
- We have way more resources contributed by the community that we are currently using on the public server. These should be public, so I cleaned them up to make sure there were no orphaned item references and pushed them here.
- Mounts.json currently fails to parse because #43 introduced a trailing comma that the deserializer doesn't accept.


yungcomputerchair:
Auto-merging since this is a data only change and needed to fix init crash.

GroaxyVRC:
Whoops forgot to remove the last comma

PR #54, opened 2026-07-08
No-op on friend req accept if already friends (fixes #49)

### LINKED ISSUE #49: Mutual friend requesting at the same time crashes the server
## Describe the bug
From Haylie:
> dev and i accidentally sent/accepted Friend Requests at the same time
> and it crashed the server i think LOL
> we all DC'd

### Steps to reproduce
- Two players send friend requests to each other
- Both players accept the requests
- Observe crash ???

### Expected behavior
no crash, whichever request is next should either no-op or return an error

yungcomputerchair:
We currently only have checks for already-friends at request time, but not accept time.


StimpyDev:
Looks fine to me

PR #55, opened 2026-07-09
Improve DB efficiency of SC shop handler

yungcomputerchair:
SC shop purchases are currently hammering the DB with a minimum of 4 database accesses per purchase (more if it's a multi-item bundle). Not only is this inefficient but it also is prone to desync with the client if one of the saves fails.

This PR refactors the SC shop purchase code to align with the Coin Shop purchase code, which does one read from the DB to get all the character's existing data and then one write with the changes applied. All the client-facing updates are deferred until after the DB commit so that the client either gets a success or a failure for all items.

### DIFF
--- a/src/Sanctuary.Gateway/Handlers/PacketBaseInGamePurchase/PacketInGamePurchasePlaceOrderPacketHandler.cs
+++ b/src/Sanctuary.Gateway/Handlers/PacketBaseInGamePurchase/PacketInGamePurchasePlaceOrderPacketHandler.cs
@@ -103,6 +103,7 @@ public static class PacketInGamePurchasePlaceOrderPacketHandler
         var dbCharacter = dbContext.Characters
             .Include(x => x.Items)
             .Include(x => x.Mounts)
+            .AsSplitQuery()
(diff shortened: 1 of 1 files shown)


StimpyDev:
This pull request only helps consistency, but loading the full Items/Mounts collections is likely slower than the old targeted queries for large inventories. Feels more like a stability tradeoff than a perf win. Prefer keeping the selective DB lookups and only adopting the pendingupdate / single-commit pattern.

yungcomputerchair:
> This pull request only helps consistency, but loading the full Items/Mounts collections is likely slower than the old targeted queries for large inventories. Feels more like a stability tradeoff than a perf win. Prefer keeping the selective DB lookups and only adopting the pendingupdate / single-commit pattern.

Stimpy please look at the PR with your human eyeballs. They will do better than your agent.

Plus, I think your agent was looking at the change before I added `AsSplitQuery()`; that was the only inefficiency in my initial change. It's the same as what we were doing before, except all of the items and mounts are fetched once instead of over several separate fetches. This is a strict improvement in perf and atomicity.

AlchemyDevelopment (approved):
Looks good to me.

PR #57, opened 2026-07-09
Fix show mounted players to other clients and some cleanup

StimpyDev:
Fixes mounted players not appearing correctly to other clients. Mount visibility now uses the same packet flow the client expects, and dismount/teleport/disconnect edge cases are handled more reliably.

### DIFF
--- a/src/Sanctuary.Game/Entities/Mount.cs
+++ b/src/Sanctuary.Game/Entities/Mount.cs
@@ -52,13 +52,8 @@ public class Mount : Npc
     {
         var packet = base.GetAddNpcPacket();
 
-        packet.RiderGuid = Rider.Guid;
+        packet.RiderGuid = 0;
 
         return packet;
-    }
-
-    public override void Dispose()
-    {
-        base.Dispose();
     }
 }
--- a/src/Sanctuary.Game/Entities/Mount.cs
+++ b/src/Sanctuary.Game/Entities/Mount.cs
@@ -56,4 +56,18 @@ public class Mount : Npc
 
         return packet;
     }
+
+    public PacketMountResponse GetMountResponsePacket()
+    {
+        return new PacketMountResponse
+        {
+            RiderGuid = Rider.Guid,
+            MountGuid = Guid,
+            Seat = Seat,
+            QueuePosition = QueuePosition,
+            Unknown = 1,
+            CompositeEffectId = 46, // PFX_Teleport_Flash
+            NameVerticalOffset = Definition.NameVerticalOffset
+        };
+    }
 }
--- a/src/Sanctuary.Game/Entities/Player.cs
+++ b/src/Sanctuary.Game/Entities/Player.cs
@@ -263,12 +263,10 @@ public sealed class Player : ClientPcData, IEntity
     {
         foreach (var npc in npcs)
         {
-            var playerUpdatePacketAddNpc = npc.GetAddNpcPacket();
+            if (npc is Mount)
+                continue;
 
-            if (npc is Mount mount && mount.Rider.Guid == Guid)
-                playerUpdatePacketAddNpc.RiderGuid = 0;
-
-            SendTunneled(playerUpdatePacketAddNpc);
+            SendTunneled(npc.GetAddNpcPacket());
         }
 
         var playerUpdatePacketNpcRelevance = new PlayerUpdatePacketNpcRelevance();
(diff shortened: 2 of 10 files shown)


yungcomputerchair:
can you write a description explaining the bugs and how these changes fix them please?

StimpyDev:
> can you write a description explaining the bugs and how these changes fix them please?

Oh my bad i forgot

src/Sanctuary.Gateway/Handlers/PacketWorldTeleportRequestHandler.cs
```diff
@@ -39,8 +39,8 @@ public static bool HandlePacket(GatewayConnection connection, Span<byte> data)
         var position = player.Position;
         var rotation = player.Rotation;
 
-        connection.Player.Mount?.UpdatePosition(position, rotation);
         connection.Player.UpdatePosition(position, rotation, updateZoneArea: false);
+        connection.Player.Mount?.UpdatePosition(position, rotation);
```
yungcomputerchair: Is there any reason not to move the mount update position inside of `Player.UpdatePosition()`? It seems fragile to repeat it after each call, especially since the mount is bound to the player

src/Sanctuary.Gateway/Handlers/PacketClientFinishedLoadingHandler.cs
```diff
@@ -24,33 +24,25 @@ public static bool HandlePacket(GatewayConnection connection)
...
-
-            connection.Player.SendTunneled(packetMountResponse);
+            connection.Player.SendTunneled(new PacketMountResponse
+            {
+                RiderGuid = connection.Player.Guid,
+                MountGuid = mount.Guid,
+                Seat = mount.Seat,
+                QueuePosition = mount.QueuePosition,
+                Unknown = 1,
+                CompositeEffectId = 46,
```
yungcomputerchair: You removed the `// PFX_Teleport_Flash` comment on the composite effect ID. can you bring it back?

src/Sanctuary.Gateway/Handlers/PacketClientFinishedLoadingHandler.cs
```diff
@@ -24,33 +24,25 @@ public static bool HandlePacket(GatewayConnection connection)
...
+                NameVerticalOffset = mount.Definition.NameVerticalOffset
```
yungcomputerchair: why was this commented out before?

src/Sanctuary.Gateway/Handlers/MountBasePacket/PacketMountSpawnHandler.cs
```diff
@@ -76,22 +76,20 @@ public static void SpawnMount(GatewayConnection connection, PacketMountInfo moun
...
-        connection.Player.SendTunneledToVisible(packetMountResponse, true);
+        connection.Player.SendTunneled(mount.GetAddNpcPacket());
+        connection.Player.SendTunneled(new PacketMountResponse
+        {
+            RiderGuid = connection.Player.Guid,
+            MountGuid = mount.Guid,
+            Seat = mount.Seat,
+            QueuePosition = mount.QueuePosition,
+            Unknown = 1,
+            CompositeEffectId = 46,
```
yungcomputerchair: comment stripped

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -310,9 +308,29 @@ public void OnAddVisiblePlayers(params IEnumerable<Player> players)
...
+                SendTunneled(new PacketMountResponse
+                {
+                    RiderGuid = player.Guid,
+                    MountGuid = player.Mount.Guid,
+                    Seat = player.Mount.Seat,
+                    QueuePosition = player.Mount.QueuePosition,
+                    Unknown = 1,
+                    CompositeEffectId = 46,
+                    NameVerticalOffset = player.Mount.Definition.NameVerticalOffset
+                });
```
yungcomputerchair: I see this response repeated in a lot of places; can we add a helper to the `Mount` class to generate it for us? Similar to `GetAddNpcPacket()`

src/Sanctuary.Game/Entities/Mount.cs
```diff
@@ -52,13 +52,8 @@ public override PlayerUpdatePacketAddNpc GetAddNpcPacket()
...
-        packet.RiderGuid = Rider.Guid;
+        packet.RiderGuid = 0;
 
         return packet;
     }
-
-    public override void Dispose()
-    {
-        base.Dispose();
-    }
```
yungcomputerchair, line 63: why was this removed?

yungcomputerchair (changes requested):

src/Sanctuary.Gateway/Handlers/PacketWorldTeleportRequestHandler.cs
StimpyDev: I'll move it into `Player.UpdatePosition()` so the mount follows automatically whenever the player moves

src/Sanctuary.Gateway/Handlers/PacketClientFinishedLoadingHandler.cs
StimpyDev: Will restore the commen, that was unintentionally dropped

src/Sanctuary.Gateway/Handlers/PacketClientFinishedLoadingHandler.cs
StimpyDev: It was commented out because the old code referenced `mountDefinition`, which doesn't exist in that handler's scope only `mount` is available. It was effectively a leftover TODO that never compiled if uncommented.

src/Sanctuary.Gateway/Handlers/MountBasePacket/PacketMountSpawnHandler.cs
StimpyDev: Same as above

src/Sanctuary.Game/Entities/Player.cs
StimpyDev: Agreed. I'll add something like `Mount.GetMountResponsePacket()`

src/Sanctuary.Game/Entities/Mount.cs
StimpyDev: The override only called `base.Dispose()` with no extra mount-specific cleanup, so it was a no-op. Removing it doesn't change behavior disposal still goes through `Npc.Dispose()`.

yungcomputerchair (approved):
Solid

StimpyDev:
People are gonna love this fix

PR #61, opened 2026-07-10
Stop assigning default title on character create

StimpyDev:
This removes the sniper title when joining for the first time


StimpyDev:
Merging this automatic

yungcomputerchair:
i'd say this is fine for small changes, but I would still discourage merging without a stamp unless there's a really good reason to

PR #63, opened 2026-07-10
Only send friend accepted after successful DB save

StimpyDev:
Stops the inviter from getting a false "friend accepted" message when the server fails to save the friendship.


StimpyDev:
Also who added `inviteeDbCharacter` is this a duplcate thing?

yungcomputerchair (approved):

yungcomputerchair:
> Also who added `inviteeDbCharacter` is this a duplcate thing?

Not sure; I would look at the blame and see where it came from if you're concerned about it

PR #66, opened 2026-07-11
Update Nuget Packages

StimpyDev:
Bump all the nuget packages to their latest versions.

### DIFF
--- a/src/Sanctuary.WebAPI/Sanctuary.WebAPI.csproj
+++ b/src/Sanctuary.WebAPI/Sanctuary.WebAPI.csproj
@@ -14,8 +14,6 @@
 
     <PackageReference Include="BCrypt.Net-Next" />
 
-    <PackageReference Include="Refit.HttpClientFactory" />
-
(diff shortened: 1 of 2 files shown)


src/Sanctuary.WebAPI/Sanctuary.WebAPI.csproj
```diff
@@ -14,6 +14,8 @@
 
     <PackageReference Include="BCrypt.Net-Next" />
 
+    <PackageReference Include="Refit.HttpClientFactory" />
```
yungcomputerchair: How come this was added?

yungcomputerchair (approved):

src/Sanctuary.WebAPI/Sanctuary.WebAPI.csproj
StimpyDev: unintentional i think, my bad i will remove it because we dont use that anyway

PR #68, opened 2026-07-12
Refactor database layer

EDITzDev:
Update solution structure to .slnx

Add basic connection test for MySql/Sqlite.
Rename Sanctuary.Database.Sq**L**ite to Sanctuary.Database.Sq**l**ite.
Remove Sanctuary.Entities project and centralize entities in Sanctuary.Database.
Centralize user secrets across projects and add a default database.json for releases.
Split database providers into MySql/Sqlite projects with dedicated contexts, factories and extensions.


src/Sanctuary.slnx
yungcomputerchair, line 1: Big fan of switching to .slnx

src/Sanctuary.WebAPI/Sanctuary.WebAPI.csproj
```diff
@@ -22,7 +18,13 @@
...
   <ItemGroup>
     <ProjectReference Include="..\Sanctuary.Core\Sanctuary.Core.csproj" />
-    <ProjectReference Include="..\Sanctuary.Database\Sanctuary.Database.csproj" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(Configuration)'=='Release'">
+    <None Remove="NLog.Development.config" />
+    <Content Remove="NLog.Development.config" />
+    <None Remove="appsettings.Development.json" />
+    <Content Remove="appsettings.Development.json" />
```
yungcomputerchair, line 27: Can you document how to use these development files in the README?

src/Sanctuary.Database/database.json
yungcomputerchair, line 1: Good call having a default here, but can you add it to `.gitignore` as well so people don't accidentally include their DB config in their changes?

src/Sanctuary.Database.MySql/Configuration/DbUserConfiguration.cs
```diff
@@ -23,6 +23,10 @@ public void Configure(EntityTypeBuilder<DbUser> builder)
 
         builder.Property(u => u.IsMember).IsRequired().HasDefaultValue(false);
         builder.Property(u => u.IsAdmin).IsRequired().HasDefaultValue(false);
+        builder.Property(u => u.IsMod).IsRequired().HasDefaultValue(false);
+
+        builder.Property(u => u.LockedUntil).IsRequired(false);
+        builder.Property(u => u.MutedUntil).IsRequired(false);
```
yungcomputerchair, line 29: good catch that these were missing

yungcomputerchair (changes requested):
This is pretty solid, the only blocker IMO is that `database.json` should be added to the .gitignore

yungcomputerchair:
It would be cool if we had one set of migrations for both backends, but it sounds like this isn't supported (otherwise you probably would have consolidated that, too).

I wouldn't worry about updating the dockerfiles; this changes the project structure a bit, so I'll rebase my Docker PR on top after we get this in.

Can you validate that your tests ran? The test output isn't very verbose: https://github.com/Open-Source-Free-Realms/Sanctuary/actions/runs/29204888861/job/86682496232

src/Sanctuary.WebAPI/Sanctuary.WebAPI.csproj
EDITzDev: This change removes the `*.Development.*` files from release builds because they aren't used. These files are only needed in the Development environment, where they allow configuration settings to be overridden. The framework automatically sets the environment to Development during debug builds, so the files are unnecessary in release builds.

I you still want me to put something on the main README lmk.

yungcomputerchair: Change checks out but yeah we should probably update our docs at some point

src/Sanctuary.Database/database.json
EDITzDev: The plan was to include this file only in release builds so users could customize it with their preferred provider and connection string. But I forgot that someone could accidentally commit their database config to git.

I've removed it an updated the database README to include samples and align with the new project structure.

EDITzDev:
> It would be cool if we had one set of migrations for both backends, but it sounds like this isn't supported (otherwise you probably would have consolidated that, too).
> 
> I wouldn't worry about updating the dockerfiles; this changes the project structure a bit, so I'll rebase my Docker PR on top after we get this in.
> 
> Can you validate that your tests ran? The test output isn't very verbose: https://github.com/Open-Source-Free-Realms/Sanctuary/actions/runs/29204888861/job/86682496232

Yeah, unfortunately having centralised migrations for all providers isn't really possible. Each EF Core provider has its own types, capabilities, and provider-specific behaviours, so the migrations tend to be tied to the provider they were generated for.

yungcomputerchair (approved):

PR #69, opened 2026-07-14
Consistent NPC IDs

yungcomputerchair:
For things like quests, we need our NPC IDs to be consistent instead of dynamically assigned.
Luckily, `NPCs.json` already has NPC IDs defined in a GUID-agnostic domain, so we can tack on the base NPC GUID to each one as they're loaded.

For newly spawned NPCs, we can still assign IDs dynamically; the counter starts at the next GUID value starting from the last loaded GUID + 1. That way the range stays contiguous and we won't run into GUID conflicts when we eventually add NPC write-back support.


cillow19 (approved):
LGTM

PR #71, opened 2026-07-14
Remove bad profiles

yungcomputerchair:
Removes these two bad profiles from `Profiles.json`:

image


StimpyDev:
Yep, saw these too. LGTM

PR #75, opened 2026-07-15
Fix AddNpc NameColor type: int ARGB, not float

CarterW24:
The client reads AddNpc's NameColor as an int32 ARGB color (m_nNameColor); the float declaration mangles any assigned color through the int→float conversion. Both are 4 bytes on the wire and the only current assignment is `default`, so nothing breaks today — the field is just unusable.

Verified live: as an int, a static ARGB value colors the NPC's overhead nameplate as specified (tested red, purple, green). 0 keeps the client default.


yungcomputerchair (approved):
I think we straight up want to use `Color` for this: https://learn.microsoft.com/en-us/dotnet/api/system.drawing.color?view=net-10.0

but since the client does more with this value than directly translate it to a color, I'd say this is good for now

StimpyDev:
LGTM, but remove the // comments, unless we really need it for documentation.

yungcomputerchair:
> LGTM, but remove the // comments, unless we really need it for documentation.

I'd say the comment is fair for now since this value isn't very straightforward.

Issue #64, opened 2026-07-10
Admins & mods are visually indistinct from normal players


yungcomputerchair:
Currently, players that belong to admin/mod accounts are visually indistinct from normal users.
It would be ideal if each role had an exclusive colored nameplate so that players can tell who is a mod and who is an admin.

LaylsInAction:
Id like to add onnnn, they are job specific.
Enforcer, Referee, Ambassador, ect. 
So really if someone could build the jobs template, you could bestow it to certain members of staff!

cillow19:
Picking this up - will bring up any future questions.

yungcomputerchair:
> Picking this up - will bring up any future questions.

Thank you so much!!

@StimpyDev FYI

raisingkaines:
Ambassador doesn't exist as a real role in the files. It's just a reskinned referee.

cillow19:
PR for referees is open for review! Sorry for the long wait!!!!

PR #78, opened 2026-07-15
add referee profile alias for designated characters

raisingkaines:
the referee job needs some work icon wise and shows up as a second adventurer job but otherwise take my scraps

### DIFF
--- a/src/Sanctuary.Game/Entities/Player.cs
+++ b/src/Sanctuary.Game/Entities/Player.cs
@@ -20,6 +20,10 @@ namespace Sanctuary.Game.Entities;
 
 public sealed class Player : ClientPcData, IEntity
 {
+    public const int RefereeProfileId = 58;
+    public const int RefereeTitleId = 900100;
+    public const int RefereeTitleNameId = 900100;
+
     private readonly UdpConnection _connection;
     private readonly IResourceManager _resourceManager;
 
@@ -38,6 +42,7 @@ public sealed class Player : ClientPcData, IEntity
 
     public bool IsAdmin { get; set; }
     public bool IsMod { get; set; }
+    public bool IsReferee { get; set; }
     public DateTimeOffset? MutedUntil { get; set; }
 
     public ClientPcProfile ActiveProfile =>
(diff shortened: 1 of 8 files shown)


src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -38,6 +42,7 @@ public sealed class Player : ClientPcData, IEntity
 
     public bool IsAdmin { get; set; }
     public bool IsMod { get; set; }
+    public bool IsReferee { get; set; }
```
yungcomputerchair, line 45: We don't want separate state for this, should just gate access to the profile on `IsMod`

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -20,6 +20,10 @@ namespace Sanctuary.Game.Entities;
 
 public sealed class Player : ClientPcData, IEntity
 {
+    public const int RefereeProfileId = 58;
+    public const int RefereeTitleId = 900100;
+    public const int RefereeTitleNameId = 900100;
```
yungcomputerchair, line 25: We should have all of the data for the referee profile in `Profiles.json` and then have just **one** constant for the profile ID. That's a lot cleaner and more flexible than defining the entire profile in code

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -508,11 +513,12 @@ public PlayerUpdatePacketAddPc GetAddPcPacket()
             MaxMovementSpeed = Stats[CharacterStatId.MaxMovementSpeed],
 
             IsUnderage = Age < 18,
-            IsMember = MembershipStatus != 0,
+            IsMember = !IsReferee && MembershipStatus != 0,
```
yungcomputerchair, line 516: `IsMember` is a basic flag which should only be composed from the membership bit in the DB; ref status is totally disjoint

src/Sanctuary.Gateway/GatewayConnection.cs
```diff
@@ -426,6 +433,78 @@ public bool CreatePlayerFromDatabase(DbCharacter dbCharacter)
         return true;
     }
 
+    private static bool IsRefereeCharacter(DbCharacter character)
+    {
+        return HasName(character, "Raising", "Kaines")
+            || HasName(character, "Jags")
+            || HasName(character, "Layla")
+            || HasName(character, "Haylie");
```
yungcomputerchair, line 441: Hardcoded names....

yungcomputerchair (changes requested):
I appreciate you submitting this but I don't think this is a very flexible way of implementing this feature. The referee profile should just be a normal profile that is injected into the profile list sent to moderator characters. There's a lot of unnecessary overhead here.

Also I believe @cillow19 claimed this in #64; not sure what her implementation will look like but I would spend time working on something other than this unless she doesn't plan on finishing it.

cillow19:
> I appreciate you submitting this but I don't think this is a very flexible way of implementing this feature. The referee profile should just be a normal profile that is injected into the profile list sent to moderator characters. There's a lot of unnecessary overhead here.
> 
> Also I believe @cillow19 claimed this in #64; not sure what her implementation will look like but I would spend time working on something other than this unless she doesn't plan on finishing it.

This is correct. I am working on this issue and will have a PR up soon... taking some time because work is kicking my butt lol and I've been thinking about how the command registry will be impacted. But we can discuss that when my PR is up.

PR #79, opened 2026-07-16
fix: allow reuse of empty guild names

raisingkaines:
the title is explanatory


yungcomputerchair (approved):
The comment is a little verbose but I think it's warranted here.

PR #80, opened 2026-07-16
Whitelist-Only Complete Enforcer Job Profile

raisingkaines:
Adds the whitelist-only Enforcer job profile, including its custom appearance, red nameplate, actor loading support, and walk, run, and backpedal animations. All Enforcer-specific changes are isolated for clean integration.
645723b062b3720b92a6bc8bd76d0103


src/Sanctuary.Gateway/GatewayConnection.cs
```diff
@@ -27,6 +27,14 @@ namespace Sanctuary.Gateway;
 
 public class GatewayConnection : UdpConnection
 {
+    public const int EnforcerProfileId = 20;
+    public const int EnforcerModelId = 536;
+    public const int EnforcerProfileNameId = 2301;
+    private const int EnforcerJacketDefinitionId = 309;
+    private const int EnforcerPantsDefinitionId = 310;
+    private const int EnforcerJacketSlot = 3;
+    private const int EnforcerPantsSlot = 4;
```
yungcomputerchair, line 36: We absolutely do not want random game-related constants in this file, it's totally the wrong layer (transport).. also, these slot numbers aren't specific to the Enforcer role

src/Sanctuary.Core/Configuration/GatewayServerOptions.cs
```diff
@@ -25,4 +25,12 @@ public sealed class GatewayServerOptions : ServerOptions
...
\ No newline at end of file
+
+    public EnforcerWhitelistEntry[] EnforcerWhitelist { get; set; } = [];
+}
+
+public sealed class EnforcerWhitelistEntry
+{
+    public ulong UserId { get; set; }
+    public ulong CharacterId { get; set; }
+}
```
yungcomputerchair, line 36: We definitely don't need an entirely separate allowlist for this, and I don't think we want it, either. We already have flags in the DB for whether a character has admin or mod privileges; we should gate on that instead of having an entire additional, unrelated registry that is scoped to just this role.

src/Sanctuary.Gateway/Handlers/BaseInventoryPacket/InventoryPacketEquipByGuidHandler.cs
```diff
@@ -40,6 +40,15 @@ public static bool HandlePacket(GatewayConnection connection, ReadOnlySpan<byte>
...
 
+        if (connection.IsEnforcer && connection.Player.ActiveProfileId == GatewayConnection.EnforcerProfileId)
+        {
+            _logger.LogInformation(
+                "Ignored an equip request while the Enforcer runtime profile is active. ( Player: {PlayerGuid}, Item: {ItemGuid} )",
+                connection.Player.Guid,
+                packet.Guid);
+            return true;
+        }
+
```
yungcomputerchair, line 51: Why do we need these guards everywhere? It's a lot of bloat for something that doesn't seem likely to happen (or cause problems if it does)

yungcomputerchair (changes requested):
I have no idea what's going on with all these deployment files you checked in; they seem to be mostly AI testing infra which we definitely don't want. All the documentation also seems to be AI-written RE/testing notes that are mostly unrelated to the feature.

For gating the feature we don't need additional infrastructure like an allowlist, we can just give it to mods (already tracked in DB).

The handler code seems somewhat usable but it still needs some tweaking.

I'll leave this open for you to clean up since there's some usable stuff in here, but this is a clear case of "just because it works doesn't mean it's good, maintainable code". This is way too much crud for such a small feature.

PR #81, opened 2026-07-17
Fix GUID spoof on move/jump, rename guard, and tell echo on ignore

StimpyDev:
Movement/jump now always broadcast the real player GUID, rename rejects GUID mismatches, and ignored tells no longer show an echo to the sender.


yungcomputerchair (changes requested):
What's the motive behind the login change? We should absolutely still burn the session as soon as it's used. That ticket is one-use for a reason: if the packet is captured, it can be replayed. That's a security risk whether you're an admin or not.

The rest looks good

StimpyDev:
> What's the motive behind the login change? We should absolutely still burn the session as soon as it's used. That ticket is one-use for a reason: if the packet is captured, it can be replayed. That's a security risk whether you're an admin or not.
> 
> The rest looks good

Yeah fair point i was thinking about the ux side (having to log in again after a lock reject), but you’re right that the ticket has to be one use either way. If we leave it reusable that’s a replay risk. I’ll put the session burn back before the lock check.

yungcomputerchair (approved):

PR #82, opened 2026-07-17
Keep pending friend invites until DB accept succeeds and block ignoring friends via interaction

StimpyDev:
Keeps friend invites until accept is saved successfully, and blocks ignoring friends from the interaction menu, which was missing

### DIFF
--- a/src/Sanctuary.Game/Entities/Player.cs
+++ b/src/Sanctuary.Game/Entities/Player.cs
@@ -395,7 +395,7 @@ public sealed class Player : ClientPcData, IEntity
         {
             commandPacketInteractionList.List.Interactions.Add(StopIgnoringInteraction.Data);
         }
-        else
+        else if (!Friends.Any(x => x.Guid == player.Guid))
         {
             commandPacketInteractionList.List.Interactions.Add(IgnoreInteraction.Data);
         }
--- a/src/Sanctuary.Game/Interactions/IgnoreInteraction.cs
+++ b/src/Sanctuary.Game/Interactions/IgnoreInteraction.cs
@@ -37,6 +37,9 @@ public class IgnoreInteraction : IInteraction
         if (player.Ignores.Any(x => x.Guid == otherPlayer.Guid))
             return;
 
+        if (player.Friends.Any(x => x.Guid == otherPlayer.Guid))
+            return;
+
         using var dbContext = _dbContextFactory.CreateDbContext();
 
         var dbCharacter = dbContext.Characters.FirstOrDefault(x => x.Id == GuidHelper.GetPlayerId(player.Guid));
--- a/src/Sanctuary.Gateway/Handlers/BaseCommandPacket/CommandPacketConfirmFriendResponseHandler.cs
+++ b/src/Sanctuary.Gateway/Handlers/BaseCommandPacket/CommandPacketConfirmFriendResponseHandler.cs
@@ -42,7 +42,7 @@ public static class CommandPacketConfirmFriendResponseHandler
 
         _logger.LogTrace("Received {name} packet. ( {packet} )", nameof(CommandPacketConfirmFriendResponse), packet);
 
-        if (!connection.Player.IncomingFriendRequests.TryRemove(packet.Guid))
+        if (!connection.Player.IncomingFriendRequests.Contains(packet.Guid))
             return true;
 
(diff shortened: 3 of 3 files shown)


src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -395,7 +395,7 @@ public void OnInteract(Player player)
         {
             commandPacketInteractionList.List.Interactions.Add(StopIgnoringInteraction.Data);
         }
-        else
+        else if (!Friends.Any(x => x.Guid == player.Guid))
```
yungcomputerchair, line 398: This seems like a bandaid in the wrong place. It shouldn't be possible to have someone on both your friend *and* ignore lists. I would keep this check simple and if this state is possible we gotta fix it in the right place. The codebase gets verbose if we have to do this check everywhere; the dev should be able to assume that if someone is on the ignore list, they can't be a friend.

src/Sanctuary.Game/Interactions/IgnoreInteraction.cs
```diff
@@ -37,6 +37,9 @@ public void OnInteract(Player player, IEntity other)
         if (player.Ignores.Any(x => x.Guid == otherPlayer.Guid))
             return;
 
+        if (player.Friends.Any(x => x.Guid == otherPlayer.Guid))
```
yungcomputerchair, line 40: Same as above, though this check is not as verbose

src/Sanctuary.Gateway/Handlers/BaseCommandPacket/CommandPacketConfirmFriendResponseHandler.cs
```diff
@@ -54,26 +54,36 @@ public static bool HandlePacket(GatewayConnection connection, ReadOnlySpan<byte>
...
                 friendMessagePacket.Type = FriendMessageType.FriendAddRequestAccepted;
                 break;
 
             case 1:
+                connection.Player.IncomingFriendRequests.TryRemove(packet.Guid);
                 friendMessagePacket.Type = FriendMessageType.FriendAddRequestDeclined;
                 break;
 
             case 2:
+                connection.Player.IncomingFriendRequests.TryRemove(packet.Guid);
```
yungcomputerchair, line 76: Not a fan of the repetition here.. it's the same line of code copy-pasted for each case. Can you rewrite so the remove only happens once, after the switch?

src/Sanctuary.Gateway/Handlers/BaseCommandPacket/CommandPacketConfirmFriendResponseHandler.cs
```diff
@@ -54,26 +54,36 @@ public static bool HandlePacket(GatewayConnection connection, ReadOnlySpan<byte>
...
                 friendMessagePacket.Type = FriendMessageType.FriendAddRequestTimedOut;
                 break;
 
             default:
                 break;
         }
 
+        if (friendMessagePacket.Type == default)
+            return true;
+
```
yungcomputerchair, line 86: I think it would be cleaner to just handle this in the `default` case, since the switch cases are what set the type.
```suggestion
            default:
                return true;
         }
```

yungcomputerchair (changes requested):

src/Sanctuary.Game/Entities/Player.cs
StimpyDev: Agreed. Friends/ignore exclusivity is enforced on accept and when ignoring. This just hides ignore for friends so the menu isn’t a no-op, happy to drop it if you prefer.

src/Sanctuary.Game/Interactions/IgnoreInteraction.cs
StimpyDev: Same rule as `CommandPacketIgnoreRequestHandler` can’t ignore a friend. Can remove if you’d rather keep that only there.

src/Sanctuary.Game/Entities/Player.cs
yungcomputerchair: Oh, nah, this makes sense then. Forgot there's a visual component to this.

StimpyDev:
Will fix the remaining changes when i'm back from vacation!

yungcomputerchair (approved):
Looks good, thanks. Sorry for the delayed review

PR #86, opened 2026-07-18
Implement Collections

amuralle:
## Summary

- Add capture-validated collection and collection-entry serialization to player state.
- Restore Briarwood Mushrooms as a basic MVP with common and rare interactable node pools for demonstration, weighted item drops, reward notifications (a stand-in for when I eventually figure out how the collection notification packet looks), despawn (now using the remove gracefully packet), and randomized respawn.
- Refresh an open Collections panel immediately after pickup using the validated authoritative self packet - this took me days.
- Add admin-only commands for persistent hardpoint placement, removal, listing, and live pool tuning in-line with some other admin commands.
- Add 76 runtime-validated Briarwood hardpoints plus focused resource and pool-policy tests that mimic an in-game state that players can go try out - I think they'd do backflips for this on the live server.

## Why this scope

Briarwood Mushrooms was the narrowest collection with both a preserved player-state row and repeatedly validated client visuals- a few of the other ones have icons and names that are unclear to me at the moment. This establishes one complete collection without guessing metadata for unrelated collection families, but I've got maybe 2-3 more in the chamber I'll roll out incrementally as they get complete. I did however figure out most if not all of the nodes, and I'll attach them in a comment, that part was easy.

Adventure coins are my priority for follow-up work since they're fan favorites but also because they are structured a little differently than other collections. Their models and historical placements are known, but non-inventory collection progress needs a per-character database representation before those nodes can be implemented safely. The player-state builder contains a focused TODO at that extension point.

## Architecture related stuff

- `Collections.json` maps the captured client collection row to inventory item definitions.
- `CollectionNodeTypes.json` owns visuals, interaction metadata, and weighted drop tables.
- `CollectionNodePools.json` owns zone, active-count, and cooldown policy.
- `CollectionNodeSpawns/<ZoneDefinitionId>/<PoolKey>.json` stores stable admin-authored hardpoints in per-zone, per-pool files. The path supplies zone and pool ownership, so each row only stores its ID, position, and heading, and authoring rewrites only the affected pool.
- `CollectionNode` provides single-winner interaction reservation and asks its zone to complete collection atomically.
- `BaseZone` owns active node instances, pool capacity, randomized selection, and a tick-driven cooldown/refill queue.
- The Gateway validates range, persists the item grant, updates in-memory inventory, refreshes collection state, and sends the runtime-validated reward notification.
- The generic `RewardBundlePacket` serializer uses the coin shop's item-received flow as its behavioral reference and is intentionally reusable by collections, combat, quests, and other reward sources.
- The existing `ChatCommandRole.Admin` gate and audit logger protect all authoring commands.

Common and rare nodes advance the same collection while using separate collectable nodes for this example. This is also true for drop tables, hardpoint pools, active caps, and cooldowns. A refill avoids the just-collected hardpoint whenever another inactive location is available. This may not be true for live as I think there were rare-specific collections but I wanted to just POC rare nodes (and again, they're fan favorites). 

## Admin commands/command design

The collection commands are world-authoring tools for collections defined in the server resources. They require `ChatCommandRole.Admin` as the other admin commands do. This was the best way to place nodes in-game to get the placement just right.

Pools and node types are intentionally not created through chat. We'll define:

- A node type in `CollectionNodeTypes.json`, which owns the world model, nameplate, scale, effect, cursor, interaction range, placement offset, and weighted item drop table.
- A pool in `CollectionNodePools.json`, which gives that pool a lowercase key, associates it with one zone and one node type, and sets its active-node cap and respawn cooldown.

For example, the pool key `briarwood-mushrooms` refers to the `mushrooms` node type. Once those definitions have been loaded, an admin uses that pool key to author its hardpoints in the associated zone.

The node type's `DropTable` assigns a relative weight to each `ItemDefinitionId`; an item's probability is its weight divided by the sum of all weights in that node type. The committed common and rare tables currently use equal weights, while retaining the ability to tune individual items later. Matching item IDs in `Collections.json` connect those drops to collection entries and live progress. The node's world appearance is independently selected by the node type's `ModelId`. 

```text
!admin collection pools
```

Lists the pools available in the current zone, including each pool's node type, active count, hardpoint count, target capacity, and respawn time.

```text
!admin collection place briarwood-mushrooms
```

Saves a hardpoint at the admin's current position and heading in the current zone's `CollectionNodeSpawns/<ZoneDefinitionId>/<PoolKey>.json` file. The node type's `PlacementYOffset` adjusts the vertical position to keep the model grounded. The hardpoint activates immediately when its pool has capacity; otherwise it remains saved as an inactive candidate for randomized refill. Steep terrain still requires visual judgment while placing.

```text
!admin collection configure briarwood-mushrooms 12 60
```

Persistently sets the pool to at most 12 active nodes and a 60-second post-collection cooldown, then reconciles the live zone immediately. `MaxActiveNodes = 0` activates every hardpoint and is useful while authoring. The committed player-facing defaults are 12 common nodes with a 60-second cooldown and 2 rare nodes with a 300-second cooldown.

```text
!admin collection list
!admin collection list briarwood-mushrooms
!admin collection list briarwood-mushrooms 2
```

Lists saved hardpoints in the current zone, ten per page, with stable IDs, pool names, coordinates, and active/inactive state.

```text
!admin collection remove
!admin collection remove 25
!admin collection remove #17
```

Removes the nearest persistent hardpoint using the default 10-unit radius, a supplied radius from 0 to 100 units, or an exact hardpoint ID. The pool is reconciled after removal.

**NOTE:** the commands - especially the placing commands for collection nodes, seem to trigger similar and/or repeated message filtering, which appears client side (unless I'm missing something), usually just go at a steady pace for placement and the commands will hit the server just fine.

## Review feedback addressed

I went through each review comment against the updated diff and made the following changes:

- I kept the change in `PlayerUpdatePacketNpcRelevance.cs` because it is part of the collection interaction path. Collection nodes are represented as NPCs, and this packet position carries the `HasCursor` value that makes their interaction cursor appear. I renamed the fields to match their observed wire meaning, set them from `Player.cs`, and confirmed the cursor and pickup behavior in-game.
- I removed `RewardBundlePacketTests.cs`, `ClientCollectionTests.cs`, and `PlayerUpdatePacketNpcRelevanceTests.cs`, along with the two newly added packet-test projects. I agree that those tests were primarily retesting the shared serialization interface rather than useful collection behavior.
- I replaced the fire-and-forget delayed respawn work in `BaseZone.cs` with a `PriorityQueue` that is processed from the existing zone tick. Collection respawns no longer create per-node delayed tasks, and state changes stay within the zone update path.
- I moved collection authoring and mutation behind new APIs in `IZone.cs` and `BaseZone.cs`. `ChatCommandRegistry.cs` now asks the zone to place, configure, list, or remove nodes instead of mutating the resources itself, so those operations happen atomically under `_collectionNodeLock`.
- I removed `CommittedCollectionNodeResourceTests.cs` and its resource-copy setup from `Sanctuary.Game.Tests.csproj`. The committed JSON can evolve without forcing maintenance of tests full of resource-specific counts and IDs.
- I retained the focused watcher coverage in `ResourceManagerTests.cs`, but renamed it to describe the actual failure mode: an atomic save can move its temporary file before an already queued watcher callback runs. The test protects that save/reload behavior rather than validating a particular committed resource file.
- I added exception documentation to the throwing methods in `CollectionNodePoolDefinition.cs` and `CollectionNodeTypeDefinition.cs`.
- I simplified `CollectionNodePoolDefinition.cs` to use `Random.Shared` instead of accepting an RNG parameter.
- I cleaned up the unrelated import formatting in `CommandPacketInteractRequestHandler.cs`.
- I restored `PacketChatHandler.cs` to the upstream version, removing the whitespace and file-ending-only diffs.

I also runtime-tested the revised architecture rather than relying only on unit tests. With a temporary five-second cooldown, I collected the same common node twice 6.44 seconds apart; it disappeared through the graceful-removal path, returned through the tick-driven queue, and produced no Gateway, Login, or WebAPI errors. I separately tested the client-facing removal transition with `Animate` enabled, a one-second duration, and no composite effect; the node now animates away cleanly without the unrelated mount-removal effect. I then restored the committed pool configuration of 12 active common nodes with a 60-second cooldown.
## Validation

- Isolated solution build: 0 warnings, 0 errors.
- `Sanctuary.Game.Tests`: 24 passed.
- `Sanctuary.UdpLibrary.Tests`: 1 passed.
- SQLite database test: 1 passed.
- Changed-file whitespace verification passed.
- Resource audit: 76 unique hardpoints, including 66 common and 10 rare, with no points within 10 world units of one another.

Runtime validation confirmed normal and rare visuals, interaction cursor and range, weighted item grants, reward notification, immediate collection-panel progress, despawn, cooldown, and alternate-hardpoint respawn.

Post-review runtime validation also confirmed the tick-driven refill queue: with all common hardpoints active and a temporary 5-second cooldown, the same node was collected twice 6.44 seconds apart. It gracefully disappeared, returned at the same hardpoint, and produced no Gateway, Login, or WebAPI errors. The pool was then restored to its committed 12-active / 60-second configuration.

### Latest upstream compatibility

After `Lua Engine + Zone Scripts` (#88) landed, I merged the latest `main` into this branch and resolved the overlapping zone lifecycle changes by activating collection pools from `BaseZone.OnStart()` alongside the new Lua startup hook. I retained the new optional-GUID NPC creation APIs and removed the old bulk NPC spawn path that Lua now replaces.

I rebuilt the full merged solution with 0 warnings and 0 errors, then ran 24 `Sanctuary.Game.Tests`, 2 `Sanctuary.Scripting.Tests`, and the SQLite database test successfully. I also booted the merged services and tested in-game: the Lua engine initialized, the expected 14 production collection nodes activated, ordinary world NPCs loaded normally, two pickups completed under a temporary five-second cooldown, live progress and animated removal still worked, and all Gateway, Login, and WebAPI error logs remained empty. I restored the committed `12 / 60s` common-pool configuration afterward.
## Known limitations

- Only Briarwood Mushrooms is enabled.
- Collection completion rewards are displayed from captured metadata but are not granted yet. 
- Inventory-backed collection items are retained so persisted progress is not lost - eventually I'd like to consume the first copy of that uncollected object and store it at the user level- maybe implementing coins will help with this.
- Live refresh resends authoritative player state until the exact collection delta packet is decoded. This one is a big pain and I know it can be better.
- Adventure coins and other non-inventory collections require persisted per-character progress. I think there's a clean way to do this but we'll see.


amuralle:
forgot to mention: 

### What players can currently experience

- Briarwood Mushrooms appears in the **Flora** section of the Collections panel.
- Common and rare mushroom nodes can appear at authored hardpoints around Briarwood, Bristlewood, and Briarwood Fort.
- Both node variants display their intended model, nameplate, interaction cursor, interaction range, and rarity effect.
- Interacting with a node grants one item selected from that node type's weighted drop table.
- The granted item is persisted to inventory and advances the corresponding collection entry.
- An already-open Collections panel refreshes immediately after a successful pickup.
- The client displays the item-received notification used by the existing coin shop reward flow.
- The collected node is removed with the graceful-removal packet and later replaced at a randomly selected inactive hardpoint.

The committed world data contains **76 validated hardpoints**: **66 common** and **10 rare**. The player-facing defaults keep **12 common nodes** active with a **60-second cooldown** and **2 rare nodes** active with a **300-second cooldown**. When another inactive hardpoint is available, a refill avoids immediately selecting the location that was just collected. This will give people a fun little gameplay experience in that zone.

src/Sanctuary.Packet/BasePlayerUpdatePacket/PlayerUpdatePacketNpcRelevance.cs
yungcomputerchair, line 1: Are the diffs in this file related to the feature?

src/Sanctuary.Packet.Tests/RewardBundlePacketTests.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Sanctuary.Packet.Tests;
+
+[TestClass]
+public sealed class RewardBundlePacketTests
+{
+    [TestMethod]
+    public void Serialize_WritesRewardAndEntryMetadata()
```
yungcomputerchair: This might just be my opinion, but I don't think we need serialization tests for individual packets. What value does this test actually provide? It's really just testing the write implementation for the packet, which is a general interface and not really specific to the reward packet itself.

src/Sanctuary.Game.Tests/CollectionNodePoolDefinitionTests.cs
yungcomputerchair, line 1: I really like these tests. Good job.

src/Sanctuary.Game/Zones/BaseZone.cs
```diff
@@ -157,6 +159,154 @@ public bool TryCreateNpc(NpcDefinition definition, [MaybeNullWhen(false)] out Np
...
+            {
+                if (TryCreateCollectionNode(typeDefinition, poolDefinition, spawn, out _))
+                    activated++;
+            }
+
+            return activated;
+        }
+    }
+
+    private async Task RefillCollectionNodePoolAsync(string poolKey, int collectedHardPointId)
```
yungcomputerchair: I am concerned about this. Even though the server isn’t strictly single-threaded, I think we’ve generally tried to avoid autonomous background tasks/async code and instead drive timed behavior from the zone/game loop, because it makes what the server is doing in the background a lot easier to predict. Also this is currently fire-and-forget, so any exceptions that arise from the task are unhandled.

Could you instead have a tick-driven respawn queue or something that is ticked in the zone tick? That would keep timing and state changes inside the existing update path and avoid accumulating per-node delayed tasks.

src/Sanctuary.Game.Tests/CommittedCollectionNodeResourceTests.cs
```diff
@@ -0,0 +1,61 @@
...
+
+using Sanctuary.Game.Resources;
+
+namespace Sanctuary.Game.Tests;
+
+[TestClass]
+public sealed class CommittedCollectionNodeResourceTests
+{
+    [TestMethod]
+    public void BriarwoodResources_HaveValidatedPlayerFacingConfiguration()
```
yungcomputerchair: Little concerned about the maintainability of this test. Won't the magic numbers in it change if the collection JSONs are updated? Keeping it in sync sounds like a pain. I think we generally want to avoid having tests that run against the resource files.

src/Sanctuary.Game.Tests/ResourceManagerTests.cs
```diff
@@ -0,0 +1,84 @@
...
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Sanctuary.Game.Tests;
+
+[TestClass]
+public sealed class ResourceManagerTests
+{
+    [TestMethod]
+    public void FileSystemWatcherChange_IgnoresDeletedTemporaryFile()
```
yungcomputerchair: What's the motivation behind this test?

src/Sanctuary.Game/Resources/Definitions/CollectionNodePoolDefinition.cs
```diff
@@ -0,0 +1,56 @@
...
+
+public sealed class CollectionNodePoolDefinition
+{
+    public required string Key { get; set; }
+    public int ZoneDefinitionId { get; set; }
+    public required string NodeType { get; set; }
+    public int MaxActiveNodes { get; set; }
+    public int RespawnSeconds { get; set; } = 30;
+
+    public int GetTargetActiveCount(int hardPointCount)
```
yungcomputerchair, line 19: I would introduce doc comments for this (and the below function) noting that they can throw exceptions. Same goes for any new functions that throw (not sure if there are others).

src/Sanctuary.Game/Resources/Definitions/CollectionNodePoolDefinition.cs
```diff
@@ -0,0 +1,56 @@
...
+        if (hardPointCount < 0)
+            throw new ArgumentOutOfRangeException(nameof(hardPointCount));
+
+        // A zero cap enables every hard point while a pool is being authored.
+        return MaxActiveNodes == 0 ? hardPointCount : Math.Min(MaxActiveNodes, hardPointCount);
+    }
+
+    public IReadOnlyList<CollectionNodeSpawnDefinition> SelectSpawnsToActivate(
+        IEnumerable<CollectionNodeSpawnDefinition> hardPoints, IReadOnlySet<int> activeHardPointIds,
+        int maximumToActivate, Random random, int? avoidHardPointId = null)
```
yungcomputerchair: Does `random` really need to be passed in here or can we acquire an RNG reference statically? One less parameter would be nice

src/Sanctuary.Gateway/Handlers/BaseCommandPacket/CommandPacketInteractRequestHandler.cs
```diff
@@ -1,9 +1,19 @@
-﻿using System;
+using System;
```
yungcomputerchair: nit: lots of false diffs near the imports in this PR, probably because of editor auto-formatting

src/Sanctuary.Gateway/Handlers/PacketChat/ChatCommandRegistry.cs
```diff
@@ -361,17 +368,285 @@ private static void Help(GatewayConnection connection, string[] args)
...
+        var radius = 10f;
+
+        if (args.Length > 0 && (!float.TryParse(args[0], out radius) || radius <= 0 || radius > 100))
+        {
+            SendSystemMessage(connection, "Removal radius must be between 0 and 100.");
+            return;
+        }
+
+        var playerPosition = connection.Player.Position;
+        var node = connection.Player.Zone.Npcs
```
yungcomputerchair: I think there's a race here, because you're not taking the zone's `_collectionNodeLock`. You should really hoist this logic to a new API inside `IZone` so that it can atomically mutate the collection nodes under the lock.

src/Sanctuary.Packet.Common.Tests/ClientCollectionTests.cs
```diff
@@ -0,0 +1,70 @@
...
+        "00003D0F00000000000070090000320000000B00000000000000000000000000000000E4150000080000002900000029" +
+        "000000010000000A000000A04200004C0800006300000000000000012A0000002A000000020000000A000000A1420000" +
+        "4C0800006000000000000000002B0000002B000000030000000A000000A24200004C0800007000000000000000012C00" +
+        "00002C000000040000000A000000A34200004C0800006400000000000000002D0000002D000000050000000A000000A4" +
+        "4200004C0800005C00000000000000012E0000002E000000060000000A000000A54200004C0800006800000000000000" +
+        "002F0000002F000000070000000A000000A64200004C0800006D00000000000000013000000030000000080000000A00" +
+        "0000A74200004C080000650000000000000000";
+
+    [TestMethod]
+    public void Serialize_MatchesCapturedBriarwoodRow()
```
yungcomputerchair: see other comment, not sure we need this

src/Sanctuary.Gateway/Handlers/PacketChat/PacketChatHandler.cs
```diff
@@ -67,7 +68,7 @@ public static bool HandlePacket(GatewayConnection connection, ReadOnlySpan<byte>
             _logger.LogWarning("Received {name} packet with null message. ( {packet} )", nameof(PacketChat), packet);
             return false;
         }
-
```
yungcomputerchair, line 70: nit: whitespace

src/Sanctuary.Gateway/Handlers/PacketChat/PacketChatHandler.cs
```diff
@@ -182,4 +183,4 @@ public static bool HandlePacket(GatewayConnection connection, ReadOnlySpan<byte>
 
         return true;
     }
-}
\ No newline at end of file
+}
```
yungcomputerchair: nit: file ending change

src/Sanctuary.Packet.Tests/PlayerUpdatePacketNpcRelevanceTests.cs
```diff
@@ -0,0 +1,26 @@
...
+
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Sanctuary.Packet.Tests;
+
+[TestClass]
+public sealed class PlayerUpdatePacketNpcRelevanceTests
+{
+    [TestMethod]
+    public void Serialize_MatchesCapturedCursorEntry()
```
yungcomputerchair: see other comment, not sure we need this.
Also is this related to the feature?

yungcomputerchair (changes requested):
The collections system in general seems pretty clean. I'd say the biggest sores are:
- Async fire-and-forget for collection respawning should be tick-driven instead
- Race due to the admin commands not holding the collection lock needs to be fixed
- Some of the tests are questionable

Also there are some changes in here about NPC relevance. Not sure if those were intended for this PR or a separate one. They're small, so we can probably bundle them in, but just figured I'd point it out.

Thanks for the work thus far!

src/Sanctuary.Game.Tests/CommittedCollectionNodeResourceTests.cs
amuralle: Yeah this was honestly leftover when I was unsure why some of my hardpoint placements weren't working (it was because of the client refusing to send commands spammed a bit too quickly to the gateway) - very redundant now, I will remove.

src/Sanctuary.Packet/BasePlayerUpdatePacket/PlayerUpdatePacketNpcRelevance.cs
amuralle: Agreed that declaration order itself doesn’t affect serialization; I changed the names/order to mirror the corrected wire layout. The functional change is in the packet writes. That correction made the interaction cursor appear and work for collection nodes, which are represented as NPC entities with the node mesh. It should apply to other NPC interactions that use CursorId, although I’ve only runtime-tested collection nodes so far. I assume this is how questing-interacts are done.

src/Sanctuary.Packet.Tests/RewardBundlePacketTests.cs
amuralle: That makes sense- Runtime testing provides better evidence here, so I’ll remove it

src/Sanctuary.Game.Tests/CollectionNodePoolDefinitionTests.cs
amuralle: tyvm

src/Sanctuary.Game/Zones/BaseZone.cs
amuralle: Yeah, that makes sense. I’ll replace the delayed background task with a respawn queue processed by the existing zone tick.

src/Sanctuary.Gateway/Handlers/PacketChat/ChatCommandRegistry.cs
amuralle: my eyeballs are coming out of my head with this one, I definitely see it. Thanks for the IZone tip.

amuralle:
changes are in!

yungcomputerchair (approved):
Looks good to me..

One note about `_collectionNodeLock`; I noticed that it's being taken recursively in some cases, e.g. `TryRemoveNearestCollectionNodeSpawn` which calls `TryRemoveCollectionNodeSpawn`. Technically this is fine and won't deadlock (I would definitely test to double-check) because the `lock` on `object` is re-entrant. But if the lock type ever changes to something *not* re-entrant, then this will deadlock. Restructuring to avoid the double-lock might be a pain though and isn't strictly necessary. Up to you whether you want to put a comment or something next to the lock about this (I would).

Note for myself: `_nextNpcGuid` needs to be made atomic. This is a pre-existing bug not introduced by your PR, just something I noticed that needs to be fixed since admin commands run on the gateway thread and not the zone ticker.

amuralle:
I have another PR in the chamber actually that can address this, so I'll add it there.

Issue #73, opened 2026-07-15
CI should also run apps after build to ensure successful startup


yungcomputerchair:
Pretty straightforward, we currently don't catch things like malformed resource or config files in CI because we only test for a successful build.

Ideally, after the build, we run:
- the WebAPI server
- the login server
- the gateway server
and ensure successful startup across all three.

PR #87, opened 2026-07-21
Fix DB init failure + adjust DB tests to catch similar failures

yungcomputerchair:
- DB init was failing due to `DatabaseContext` being abstract. Tests didn't catch it because they initialize the DB differently. Change the tests so that they initialize in the same way as the server so they catch init failure. (thanks @EDITzDev)
- Use a singleton in each DB provider instead of DbContextFactory, which fixes the failure.
  - This allows for `DatabaseContext` to continue being an abstract class.
  - This change makes the *factory* the singleton; DB contexts are still created fresh, so there's no risk of stale EF caches.


cillow19 (approved):

yungcomputerchair:
Implementing #73 will help catch similar breaks in the future

PR #88, opened 2026-07-22
Lua Engine + Zone Scripts

yungcomputerchair:
This PR adds a Lua engine (powered by [Lua-CSharp](https://github.com/nuskey8/Lua-CSharp), MIT) that will pave the way for us to add scripting to zones, NPCs, and more.

- All LuaCSharp access is restricted to the new `Sanctuary.Scripting` project; no Lua implementation details are exposed to the gateway.
- The project provides a `ScriptManager` that implements `IScriptManager` and is initialized through DI like the other managers.
  - On init, the manager initializes the Lua standard library with some simple overrides for logging.
  - Extremely simple Lua bindings to zones are provided in `ScriptableZone`, with `IScriptZone` being the actual contract.
    - Currently, you can access the zone ID and name, and also spawn NPCs with static or dynamic GUIDs
  - All zones that extend `BaseZone` load their respective Lua script (by name) and execute the `onStart` function at the onset of the `OnStart` event.
    - This event apparently existed before (hinted by a comment) but was removed. It felt perfect to bring back for this.
- Scripts are kept in a `Scripts` folder next to `Resources`, with subfolders based on script types.
  - Currently there is only a `Zone` subfolder since that is the only scriptable object type for now. NPC AI will follow suit later.
- A testing project, `Sanctuary.Scripting.Tests`, is provided with basic tests ensuring init passes and all zone scripts can be successfully parsed and loaded by the scripting manager.
- Since NPCs are to be spawned by zone/dungeon scripts, position and rotation fields have been removed from `Npcs.json`. This file now stores pure NPC templates for instantiation.
  - Naturally, all overworld NPC spawns are now handled by `onStart` in `FabledRealms.lua`.


EDITzDev:
I think the implementation looks good overall.

One suggestion I'd like to make is to move static NPCs from `Npc.json` to the existing per-zone .json file. My worry is that static NPCs will eventually clutter the .lua scripts and make them harder to review as more get added.

So for example in `Zones\FabledRealms.json` we'd have a `StaticNpcs` list similar to what `Npcs.json` looks like with Position/Rotation.

This would also significantly reduce the number of NPCs that need to be registered in Lua, since the majority are static. By "static," I mean NPCs that don't have any interactions, quests, movement, or other behaviours they're purely decorative.

NPCs that do have interactions, quests, movement, or other behaviour still make sense to keep in Lua.

What do you think?

yungcomputerchair:
> What do you think?

Your idea was actually my initial plan for this, but there are two big reasons why I didn't go with it:
- Multiple zones will undoubtedly share NPC templates. If there are, say, two dungeons that spawn the same wolf enemy, that means two copies of a definition with the same model ID, same name ID, same name, same AI script name (eventually) etc. As we implement all the dungeons, this will be a LOT of duplication built up over time.
- One global registry for the templates separates the concerns: it keeps only the appearance data and allows for the zone scripts to dictate what appears and where. That's the declarative stuff. Meanwhile, the scripts handle the imperative stuff, like the positioning. IMO, it doesn't matter whether an NPC is static or not; either way, it has a position, and that data should be established the same way for both static and non-static NPCs. That's why I went with instantiation in the FabledRealms script: because it would be awkward to declare some positions in JSON and some in Lua.

If you still have concerns, lmk. It's worth debating to get it right.

EDITzDev:
Fair enough, just felt like static NPCs should be stored along with the zone definition, since they'll never change.

EDITzDev (approved):

yungcomputerchair:
> Fair enough, just felt like static NPCs should be stored along with the zone definition, since they'll never change.

If this ends up being unruly in the future, we'll move to that approach.

PR #91, opened 2026-07-27
First steps towards server-side pathfinding

ImAlko:
This PR should give some basic path-finding that works on `.map` files which were originally Free Realms assets that represented simple paths in the form of a graph. The important piece is the `Pathfinder` which _should_ get created once whenever a new zone is loaded. One can use this to path find between arbitrary positions in that zone.
 
There are still many, many considerations that need to be addressed (see many of the `TODO`s and `NOTE`s riddled throughout). 

Note that I have purposely removed the testing code I used to verify in game. This is because it was a janky chat parser. Will probably need to implement that the right way.  

Also, Claude was instrumental in helping me get this together. A few of the lower-level files used for loading were not touched much by myself. I had some equivalent Python tests verify that the graphs looked reasonable (+ the in-game testing). See [this video](https://www.youtube.com/watch?v=244yPqMi9qg). 

EDIT: This PR is probably still very rough. I'm going to need some guidance on the software architecture stuff, so I expect to have to make a bunch of changes.


src/Sanctuary.Game/Zones/BaseZone.cs
```diff
@@ -84,6 +87,12 @@ protected BaseZone(BaseZoneDefinition zoneDefinition, IServiceProvider servicePr
 
         Task.Factory.StartNew(UpdateEveryTickAsync, _cancellationTokenSource.Token, TaskCreationOptions.LongRunning, TaskScheduler.Default);
         Task.Factory.StartNew(UpdateEverySecondAsync, _cancellationTokenSource.Token, TaskCreationOptions.LongRunning, TaskScheduler.Default);
+        
+        // Just in case we don't actually have the `.map` file for a particular zone.
+        if (_resourceManager.Maps.TryGetValue(Name, out var mapGraph))
+            Pathfinder = new Pathfinder<MapNode>(mapGraph.Nodes, _logger);
+        else
+            Pathfinder = null;
```
yungcomputerchair: Explicit null set here is noise

src/Sanctuary.Game/Pathfinding/IPathNode.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections.Generic;
+using System.Numerics;
+
+namespace Sanctuary.Game.Pathfinding;
+
+public interface IPathNode
```
yungcomputerchair, line 6: nit: the only class that implements this interface is `MapNode`. Maybe this should just be named `IMapNode` unless you have another use in mind for the future?

src/Sanctuary.Game/Entities/Npc.cs
```diff
@@ -331,4 +389,84 @@ private void RemoveFromZone()
...
+    {
+        Goal = new MovementGoal.FixedPosition(goalPosition);
+        RecomputePath();
+    }
+
+    public void Chase(Player player)
+    {
+        // TODO: not stable
+        throw new NotSupportedException("NPC chasing is not yet ready.");
+    }
```
yungcomputerchair: I would encourage tearing out the chasing code entirely and re-introducing it once it's stable. Right now it's just noise.

src/Sanctuary.Game/Entities/Npc.cs
```diff
@@ -111,10 +128,51 @@ public virtual void OnRemoveVisiblePlayers(params IEnumerable<Player> players)
...
+        {
+            step = toTarget;
+        }
+
+        var newPosition = currentPosition + step;
+        var newOrientation = Quaternion.CreateFromYawPitchRoll(MathF.Atan2(direction.X, direction.Z), 0f, 0f);
+
+        UpdatePosition(new Vector4(newPosition, 1f), newOrientation);
+        BroadcastPosition();
     }
```
yungcomputerchair, line 132: I agree with your comment. I would hoist all this path-related update stuff into a private `UpdatePath()` function (or even have a `Path` class that sits on the NPC), and then call that from `OnUpdate()`.

src/Sanctuary.Game/Entities/Npc.cs
```diff
@@ -111,10 +128,51 @@ public virtual void OnRemoveVisiblePlayers(params IEnumerable<Player> players)
...
 
     public virtual void UpdateEverySecond()
     {
+        // TODO: How well does this scale? Do we eventually want concurrency/parallelization (didn't really
+        // look and see if something like that is already handeled at the higher-level).
+        RecomputePath();
```
yungcomputerchair: Speaking of concurrency, `UpdateEverySecond()` and `UpdateEveryTick()` actually run on separate threads, so I would double check that `RecomputePath()` is thread/concurrency safe (at a glance, I don't think it is)

src/Sanctuary.Game/Pathfinding/Pathfinder.cs
```diff
@@ -0,0 +1,228 @@
...
+    private readonly IReadOnlyDictionary<int, TNode> _nodes;
+    private readonly ILogger _logger;
+
+    public Pathfinder(IReadOnlyDictionary<int, TNode> nodes, ILogger logger)
+    {
+        this._nodes = nodes;
+        this._logger = logger;
+    }
+
+    public List<TNode> FindPath(Vector3 startPosition, Vector3 goalPosition) {
```
yungcomputerchair: Have you profiled this? I'm a little worried about memory usage and CPU cycles (since it's synchronous). A path that takes a long time to compute will cause a noticeable hitch in the executing thread, and a path that takes a lot of nodes to compute will cause spikes in the memory footprint.

Not saying this is unacceptable-- it very well may be fine-- but it's important to optimize since this will be a hot-path for sure once we have things like dungeons with a lot of moving NPCs.

src/Sanctuary.Game/Pathfinding/MapGraphLoader.cs
```diff
@@ -0,0 +1,151 @@
...
+using System.IO;
+using System.Numerics;
+
+using Microsoft.Extensions.Logging;
+
+namespace Sanctuary.Game.Pathfinding;
+
+// NOTE: This file was pretty much mostly vibe-coded and verified with some local
+// Python tests (that performed equivalent operations). I'm going to leave leave
+// this summary here because it may be useful - Alko.
```
yungcomputerchair: No need for AI disclaimers; we all use it :)

src/Sanctuary.Game/Entities/Npc.cs
```diff
@@ -111,10 +128,51 @@ public virtual void OnRemoveVisiblePlayers(params IEnumerable<Player> players)
...
+        // Feels like this is just bloating a function that is fairly importable.
+
+        if (this.Waypoints.Count == 0)
+        {
+            return;
+        }
+
+        // NOTE: I (Alko) am not super duper familiar with how game engine physics affects NPC
+        // movements. If path following is complete garbage for some reason (I doubt it will be)
+        // come back and here and break out some basic robotics algorithms.
```
yungcomputerchair: nit: I appreciate the comment but it's not helpful here

src/Sanctuary.Game/Entities/Npc.cs
```diff
@@ -70,6 +73,20 @@ public class Npc : IEntity
...
     public bool Static { get; set; }
 
+    // TODO: Is this safe, scalable and okay for many, many NPCs? I (Alko) should think
+    // more about this, but leaving it for now.
+    public Queue<Vector3> Waypoints { get; set; } = new();
+
+    private const float TickDeltaSeconds = 0.1f;
+
+    // TODO: maybe keeping this constant is a bad idea... 
+    // Could be different for all NPCs, fix this before the PR goes in.
```
yungcomputerchair: I think it would make sense for node tolerance to be a per-NPC property (maybe we can precompute it based on the model or something) but leaving it as a constant is fine for now IMO.

Speed should be a property of the path. We can have a per-NPC base speed property that movement can default to, but ultimately, you should be able to tell an NPC to go somewhere at X speed.

src/Sanctuary.Game/Entities/Npc.cs
```diff
@@ -70,6 +73,20 @@ public class Npc : IEntity
 
     public bool Static { get; set; }
 
+    // TODO: Is this safe, scalable and okay for many, many NPCs? I (Alko) should think
+    // more about this, but leaving it for now.
+    public Queue<Vector3> Waypoints { get; set; } = new();
+
+    private const float TickDeltaSeconds = 0.1f;
```
yungcomputerchair: I would add a comment documenting this constant and why you landed on .1f

yungcomputerchair (changes requested):
Some small notes and some bigger ones. I think for the most part you did a pretty good job of matching the conventions in the codebase. The new collection looks good and I like how you separated all the pathfinding related stuff into its own folder.

I can't speak on the actual pathfinding math; it looks sane at a glance. Since you have a background on robotics I'd say we can probably trust you on that :)

src/Sanctuary.Game/Pathfinding/IPathNode.cs
ImAlko: Yeah, from my navmesh stuff I may require additional information to do things like a funneling algorithm. I want the path finder to be agnostic to that (so working on the interface). I'm not sure what I'd call the `<NEW>Node` but it'd be a `IPathNode`.

NOTE: we may never add the navmesh capability since it's likely _too complex_, but just something I was planning for just in case.

src/Sanctuary.Game/Entities/Npc.cs
ImAlko: Oof interesting. Okay, I will make sure I look into this.

src/Sanctuary.Game/Pathfinding/Pathfinder.cs
ImAlko: From limited Python testing, I was seeing on the order of a few milliseconds for searches on the `FabledRealms.map` graph (more than likely the largest `.map` file), but I can double check. I'd assume it'd be quicker in C#. I can try to do some very simple profiling in case I introduced something funky when I carried it over. I'll need to look into memory usage.

ImAlko:
@yungcomputerchair saw your comments and responded to a few for now. I'll likely have time to hit all of these on Saturday if that's okay! Thanks a ton!

src/Sanctuary.Game/Entities/Npc.cs
ImAlko: Got rid of this. It was derived from the `BaseZone` values. I'm just going to grab a derived value directly now!

src/Sanctuary.Game/Pathfinding/Pathfinder.cs
ImAlko: Results from (vibecoded) profiler (not going to check this in, unless you'd like me to):

```
 PathBuilder.TryRecompute: 0.6609 ms/call avg
 ```
 
 This was with a hardcoded start of Briarheart to a goal of Pirate's Plunder in Seaside!

File: src/Sanctuary.Game/Entities/Npc.cs (hunk 5 above)
Comment (ImAlko) [reply src/Sanctuary.Game/Entities/Npc.cs]: Made a new class with locks + changed some of the logic to be more thread safe. It's going to need a review to be certain. I'm not fully happy with it, but it should be better. Also, some of these issues go away since I'm not longer doing anything in `UpdateEverySecond` (since I removed chasing behavior).

Side note: once chasing needs to implemented, would it make more sense to implement that at the higher-level, i.e., just make use of `Npc`'s `MoveTo` I added with the input position coming from a player input? Something to think about I guess.

File: src/Sanctuary.Game/Entities/Npc.cs (hunk 4 above)
Comment (ImAlko) [reply src/Sanctuary.Game/Entities/Npc.cs:132]: The logic should be a lot less bloated now. There's a new `PathBuilder` and `PathFollower`

File: src/Sanctuary.Game/Entities/Npc.cs (hunk 3 above)
Comment (ImAlko) [reply src/Sanctuary.Game/Entities/Npc.cs]: Done

File: src/Sanctuary.Game/Zones/BaseZone.cs (hunk 1 above)
Comment (ImAlko) [reply src/Sanctuary.Game/Zones/BaseZone.cs]: Should be resolved

File: src/Sanctuary.Game/Entities/Npc.cs (hunk 9 above)
Comment (ImAlko) [reply src/Sanctuary.Game/Entities/Npc.cs]: > Speed should be a property of the path. We can have a per-NPC base speed property that movement can default to, but ultimately, you should be able to tell an NPC to go somewhere at X speed.

I have went ahead and cleaned-up the hardcoded stuff, so they are just per-NPC properties. I _think_ `Speed` is also used client-side for interpolation and visualizing the movement (I would need to check this to be sure), so I'm hesitant to have two different speeds floating around.

File: src/Sanctuary.Game/Pathfinding/Pathfinder.cs (hunk 6 above)
Comment (yungcomputerchair) [reply src/Sanctuary.Game/Pathfinding/Pathfinder.cs]: Very happy with that number

File: src/Sanctuary.Game/Entities/Npc.cs (hunk 5 above)
Comment (yungcomputerchair) [reply src/Sanctuary.Game/Entities/Npc.cs]: I'll do a closer review on the concurrency later, but I agree that I don't think we need a chase function in the code and we can just use `MoveTo` as the movement API for every behavior we need.

Comment (yungcomputerchair) [reply src/Sanctuary.Game/Entities/Npc.cs]: (Although I would maybe keep the interface flexible in case we change our minds; designs like this tend to change rapidly once things like combat and more complex NPC scripts arise)

File: src/Sanctuary.Game/Entities/Npc.cs (hunk 9 above)
Comment (yungcomputerchair) [reply src/Sanctuary.Game/Entities/Npc.cs]: Yeah, we definitely need to nail down what the client expects for speed... it would be ideal not to have two different ones. Hopefully we can unify and just give the client what it wants.

File: src/Sanctuary.Game/Pathfinding/PathState.cs
Comment (yungcomputerchair) [inline src/Sanctuary.Game/Pathfinding/PathState.cs:1]: Very nice

File: src/Sanctuary.Game/Pathfinding/Pathfinder.cs [hunk 11]
```diff
@@ -0,0 +1,235 @@
...
+        int? meetingNodeId = null;
+
+
+        // NOTE: this was ported from a Python implementation that I have.
+        // Apparently, `heapq` and PriorityQueue have difference tie breaker 
+        // behavior. This shouldn't affect the validity of the solution, but if
+        // anyone ever gets their hands on the Python version for validation, it may 
+        // not be 1:1.
+        // - Alko
+        var forwardQueue = new PriorityQueue<int, float>();
```
yungcomputerchair, line 47: nit: this comment is very likely only relevant to you. If you really wanna keep it that's fine, but typically we try to use comments sparingly to explain complex things to devs who don't own the area.

src/Sanctuary.Game/Pathfinding/Pathfinder.cs
```diff
@@ -0,0 +1,235 @@
...
+    {
+        this._nodes = nodes;
+        this._logger = logger;
+    }
+
+    // Runs bi-directional A* to compute the path between two positions. 
+    // NOTE: positions need not be a part of the graph. This introduces potential
+    // limitations if there is an obstacle between the given position and its closest
+    // node. This cannot be fixed without a navmesh/more detailed structure.
+    public List<TNode> FindPath(Vector3 startPosition, Vector3 goalPosition)
```
yungcomputerchair, line 28: nit: use C# doc comment for function annotation

src/Sanctuary.Game/Pathfinding/Pathfinder.cs
```diff
@@ -0,0 +1,235 @@
...
+    {
+        var startNode = this.FindClosestNode(startPosition);
+        var goalNode = this.FindClosestNode(goalPosition);
+
+        _logger.LogDebug(
+            "Finding path from {StartPosition} to {GoalPosition} (start node {StartNodeId}, goal node {GoalNodeId})...",
+            startPosition, goalPosition, startNode.Id, goalNode.Id);
+
```
yungcomputerchair, line 36: nit: this can probably be Trace level instead of Debug level

src/Sanctuary.Game/Pathfinding/MovementGoal.cs
```diff
@@ -0,0 +1,17 @@
+using System.Numerics;
+
+
+// TODO: implement a `ChaseEntity`.
+// - Alko
```
yungcomputerchair: nit: prefer
```suggestion
// TODO Alko: implement a `ChaseEntity`.
```

src/Sanctuary.Game/Pathfinding/MapGraphLoader.cs
yungcomputerchair: that is to say, it would be preferable to delete this top part to keep the documentation concise.

src/Sanctuary.Game/Pathfinding/MovementGoal.cs
yungcomputerchair: Also, I've been thinking this over and I do think we want an actual `ChaseEntity` goal after all (I know we talked offline about omitting it and just using `MoveNpc` for everything). That way we don't need extra state + processing overhead in the NPCs.

Absolutely does not need to be in this PR. It can come later.

src/Sanctuary.Game/Pathfinding/MapGraphLoader.cs
```diff
@@ -0,0 +1,151 @@
...
+                var y = reader.ReadSingle();
+                var z = reader.ReadSingle();
+                var neighborCount = reader.ReadUInt32();
+
+                if (neighborCount > MaxSaneNeighborCount)
+                {
+                    logger.LogError(
+                        "Map file \"{Path}\" has a suspicious neighbor_count={Count} for node {Id} " +
+                        "at offset {Offset} - likely a parse desync.",
+                        filePath, neighborCount, id, stream.Position - 4);
```
yungcomputerchair, line 79: Is this likely to happen?

src/Sanctuary.Game/Entities/Npc.cs
```diff
@@ -70,6 +73,19 @@ public class Npc : IEntity
...
+    public float Speed { get; set; }
+
+    private readonly PathState _path = new();
+
+    // NOTE: This will be lazily initialized. For one, if a zone does not contain
+    // a coresponding '.map' file, it will just stay 'null'. Moreover, if 
+    // one just wants the NPC to move directly to a point with no pathfinding,
+    // this will stay 'null'. A new 'PathBuilder' is created only when pathfinding
+    // is asked for.
+    private PathBuilder? _pathBuilder;
```
yungcomputerchair, line 82: nit: this lazy-init pattern is pretty common in the codebase. You can shorten or delete the comment IMO.

src/Sanctuary.Game/Entities/Npc.cs
```diff
@@ -331,4 +356,39 @@ private void RemoveFromZone()
...
+            Rotation = Rotation,
+            State = 1,
+            Unknown = 0
+        };
+
+        foreach (var visiblePlayer in VisiblePlayers)
+        {
+            visiblePlayer.Value.SendTunneled(packet);
+        }
+    }
```
yungcomputerchair: Is there any reason this logic shouldn't just live at the end of `UpdatePosition()`?

yungcomputerchair (approved):
Looks good, just nitpicks and some small questions at this point. I'll test locally and wait for your thumbs up before merging.

src/Sanctuary.Game/Pathfinding/MapGraphLoader.cs
ImAlko: Honestly with the `map` files, probably not.

src/Sanctuary.Game/Pathfinding/Pathfinder.cs
ImAlko: Should be done!

src/Sanctuary.Game/Pathfinding/Pathfinder.cs
ImAlko: If its okay, I'd love to leave it because I _know_ I'm going to forget this haha

src/Sanctuary.Game/Pathfinding/Pathfinder.cs
ImAlko: Done (did the same with another `LogInformation` that I think would be spammy)

src/Sanctuary.Game/Pathfinding/MovementGoal.cs
ImAlko: This file actually isn't used in the current logic anymore. Pretty easy to re-implement if we need it

src/Sanctuary.Game/Entities/Npc.cs
ImAlko: Done

src/Sanctuary.Game/Entities/Npc.cs
ImAlko: `BroadcastPosition` is gone and its just all going to live in `UpdatePosition` now!

PR #96, opened 2026-07-31
Use native collection update packets

amuralle:
## Summary

I replaced the full player refresh used after collecting a node with the
client's native collection update packets.

The full `PacketSendSelfToClient` refresh updated the Collections panel, but it
also disturbed unrelated client state. Most visibly, collecting while mounted
could desync the mount and action UI. PR #94 tried to compensate by dismounting
the player server-side, but runtime testing showed that this does not reproduce
the client-side state transition of a normal dismount. This PR reverts that
workaround and removes the broad refresh that made it necessary.

## What changed

- Added `ClientUpdatePacketCollectionStart` for starting a collection.
- Added `ClientUpdatePacketCollectionAddEntry` for adding live progress.
- Corrected the decoded collection and collection-entry field meanings.
- Replaced the full self refresh and generic reward toast with native
  collection updates.
- Reverted the collection-triggered mount workaround from #94.
- Updated the Briarwood collection definition to use the corrected schema.

The client now creates its original collection started, progress, and completed
notifications from these packets.

## Packet research

I traced the `BaseClientUpdatePacket` collection handlers in the client:

- subtype `8`: collection start
- subtype `10`: collection add entry

The add-entry handler updates the client's collection manager, marks both
collection views dirty, and creates the native collection UI action. This
explains why the previous full player refresh could change the panel without
behaving like an original collection event.

For an unknown collection, the server sends subtype 8 with the serialized
collection state containing the first collected entry. For an already-started
collection, it sends subtype 10 with the newly collected entry.

## Runtime testing

- Started a previously unknown collection and received the native start/update
  notification.
- Added later entries and saw the open Collections panel update live.
- Completed a collection and received the native completion notification.
- Confirmed collection progress persisted after relogging.
- Collected while mounted without forcing a dismount or refreshing the full
  player state.
- Confirmed normal mount controls remained available after collecting.

## Automated validation

- `dotnet build src/Sanctuary.slnx --no-restore`
  - 0 warnings
  - 0 errors
- `Sanctuary.Game.Tests`
  - 24 passed
- `Sanctuary.Scripting.Tests`
  - 2 passed
- `Sanctuary.UdpLibrary.Tests`
  - 1 passed
- The SQLite database test passed. The MySQL integration test was unavailable
  locally because no MySQL service was running.

### DIFF
--- a/src/Resources/Collections.json
+++ b/src/Resources/Collections.json
@@ -1,9 +1,9 @@
 [
   {
-    "Id": 17054,
-    "CategoryId": 10,
+    "Id": 10,
+    "NameId": 17054,
+    "CategoryId": 3,
     "DescriptionId": 17055,
-    "Type": 3,
     "IconId": 2124,
     "IconTintId": 92,
     "HeaderMetadata": 9,
--- a/src/Sanctuary.Game.Tests/CollectionDefinitionTests.cs
+++ b/src/Sanctuary.Game.Tests/CollectionDefinitionTests.cs
@@ -15,6 +15,7 @@ public sealed class CollectionDefinitionTests
         var definition = new CollectionDefinition
         {
             Id = 17054,
+            NameId = 17055,
             CategoryId = 10,
             Entries =
             [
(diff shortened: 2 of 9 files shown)


amuralle:
Updated this PR with the next reasonable slice of the collections work.

This is now the packet work plus the actual collection-node wiring: node pickups no longer rely on SendSelfToClient() to refresh collection progress. Instead, new collection progress goes through the native collection start / add-entry packets, while duplicate already-collected item drops still use the reward bundle toast.

I also kept the expanded collection catalog and placement data out of this on purpose. My intent is for this PR to be the packet/schema/runtime wiring piece, then follow up separately with the bigger content/placement PR once this machinery looks good.

yungcomputerchair (approved):

PR #102, opened 2026-08-04
Quest system + "Take Me There" pathfinding

Sulphural:
### What's included

**Quest system**
- Data-driven quest engine (`QuestManager`) — adding a quest is a `Quests.json` entry, not new code
- Goal types: `TalkToNpc`, `ReachLocation`, `Collect`
- Mutual-exclusion quests (`ExcludesQuestIds`) for the two race-specific "Introduce Yourself" quests
- Full persistence: active quest, current goal, partial Collect progress, and completion all survive relog/restart (new `CharacterQuests` table)

**"Take Me There"**
- Real client packet flow (`ClientPathRequestPacket`/`ClientPathReplyPacket`, opcode 98) — genuine client-side auto-walk with the green breadcrumb trail, not server-forced movement
- Routes through the newly-merged `Pathfinder<MapNode>` (bidirectional A* over real `.map` waypoint data) instead of a proximity heuristic

18 quests, all data-driven from Quests.json. Goal types: TalkToNpc (all 18), Collect (3: Hasti's Homework Heist, Tools for Smashing, Box O' Rocks), ReachLocation (1: Welcome to Seaside).


StimpyDev:
Thanks for the pull request. For the future, stop providing those doc files, unless it really needs documentation.

As i see we dont use those anywhere, i dont think it's needed

yungcomputerchair:
Ok, here's how we're gonna deal with this one.
- First, I'm going to close the PR.
  - This is not because these changes are bad-- there's some good stuff in here-- but it's too much code for any software engineer to feasibly review. I don't want this to end up like the collections PR that has been iterated on for over a month.
  - Do **not** lose this code. Keep the branch around.
- Next, I'm going to have you wait for @amuralle to make a PR for reward bundles, and then get it merged.
  - He is already working on them since they serve as the correct backend for collections.
  - AI has screwed up a few packet layouts lately, so we are taking extra care to review packet layouts separately from features (or at least with minimal features alongside them).
  - After reward bundles are in, you can remove your implementation of them and adapt your code to use the ones in main.
- After that, I'll have you submit separate PRs-- **one by one**, i.e. wait for one to get reviewed and merged before putting up another-- composed of the code from this branch. Split it like this, in this order:
  1. The DB schema and migrations, the data model for quests, and the quest JSON loading.
  2. All the quest-related packet serializers and deserializers (**no handlers at this point**)
  3. All the dialog-related packet serializers and deserializers (handlers O.K. for this one)
  4. The notification badges for NPCs.
  5. The Quest Manager, handlers for the quest packets.
  6. The "Take me there" feature.

I understand that this feels like it will slow down the introduction of quests on the public server, but this lets us thoroughly review and test each component one by one without making a ton of iterations in a monster PR. I am actually pretty confident that it'll get us there faster.

If any other contribs have opinions on the split or architecture, feel free to ping here.

PR #104, opened 2026-08-06
Correct reward bundle item entry serialization

amuralle:
﻿## Summary

I corrected `RewardBundlePacket` to serialize the typed entry list expected by the client instead of treating one item reward as a flat packet payload.

- Added the shared reward-bundle entry fields and the item entry subtype.
- Preserved the packet's entry count and polymorphic entry structure.
- Updated the collection-node reward caller to use the player's actual item GUID.
- Kept the existing unique source GUID behavior used to display repeated notifications.

## Packet provenance

The original item-received behavior came from a known-working coin shop packet capture. I then inspected the Free Realms client binary in Ghidra to recover how the client parses reward bundles and their typed entries.

The client reads an entry count followed by a common entry structure and type-specific data. For a successful item entry, the subtype payload is the item GUID directly; there is no additional subtype boolean on the wire. I validated that recovered shape at runtime through collection-node item rewards, including repeated rewards for an item already owned by the player.

This PR intentionally implements only the runtime-validated item entry. Other reward entry types can be added separately as their client behavior and server use cases are validated.

## Validation

- Runtime-tested collection-node item rewards, including duplicate-item notifications.
- Built the full solution with zero warnings and zero errors.
- Passed 24 game tests, 2 scripting tests, 1 UDP test, and the SQLite database test.
- Verified formatting for every changed file.

The MySQL integration test was not run because no local MySQL service was available.

### DIFF
--- a/src/Sanctuary.Packet.Common/RewardBundleEntryBase.cs
+++ b/src/Sanctuary.Packet.Common/RewardBundleEntryBase.cs
@@ -1,36 +1,35 @@
 using Sanctuary.Core.IO;
-using Sanctuary.Packet.Common;
 
-namespace Sanctuary.Packet;
+namespace Sanctuary.Packet.Common;
 
-public abstract class RewardBundleEntry
+public abstract class RewardBundleEntryBase
 {
     public abstract RewardBundleEntryType Type { get; }
 
-    public bool Unknown;
+    public bool IsHidden;
     public int IconId;
-    public int IconTintId;
+    public int TintId;
     public int NameId;
     public int Quantity = 1;
     public int DefinitionId;
     public int Tint;
-    public string Text = string.Empty;
-    public int Unknown2;
-    public bool Unknown3;
+    public string Description = string.Empty;
+    public int ItemTextColor;
+    public bool MembersOnly;
 
     internal void Serialize(PacketWriter writer)
     {
         writer.Write((int)Type);
-        writer.Write(Unknown);
+        writer.Write(IsHidden);
         writer.Write(IconId);
-        writer.Write(IconTintId);
+        writer.Write(TintId);
         writer.Write(NameId);
         writer.Write(Quantity);
         writer.Write(DefinitionId);
         writer.Write(Tint);
-        writer.Write(Text);
-        writer.Write(Unknown2);
-        writer.Write(Unknown3);
+        writer.Write(Description);
+        writer.Write(ItemTextColor);
+        writer.Write(MembersOnly);
 
         SerializeData(writer);
     }
--- a/src/Sanctuary.Packet.Common/RewardBundleEntryType.cs
+++ b/src/Sanctuary.Packet.Common/RewardBundleEntryType.cs
@@ -1,0 +1,6 @@
+namespace Sanctuary.Packet.Common;
+
+public enum RewardBundleEntryType
+{
+    Item = 1
+}
--- a/src/Sanctuary.Packet.Common/RewardBundleEntryType.cs
+++ b/src/Sanctuary.Packet.Common/RewardBundleEntryType.cs
@@ -2,5 +2,21 @@ namespace Sanctuary.Packet.Common;
 
 public enum RewardBundleEntryType
 {
-    Item = 1
+    Item = 0x1,
+    Experience = 0x3,
+    QuestAdd = 0x6,
+    ProfileAdd = 0x7,
+    AbilityLine = 0x8,
+    CollectionAdd = 0xA,
+    CollectionEntryAdd = 0xB,
+    Token = 0xC,
+    PetTrickExperience = 0xD,
+    Recipe = 0xE,
+    ZoneFlag = 0xF,
+    CharacterFlag = 0x11,
+    WheelSpin = 0x12,
+    Trophy = 0x13,
+    ClientExitUrl = 0x14,
+    FactoryExperience = 0x16,
+    Chest = 0x17
 }
--- a/src/Sanctuary.Packet/BaseRewardPacket/RewardBundleEntry.cs
+++ b/src/Sanctuary.Packet/BaseRewardPacket/RewardBundleEntry.cs
@@ -1,0 +1,39 @@
+using Sanctuary.Core.IO;
+using Sanctuary.Packet.Common;
+
+namespace Sanctuary.Packet;
+
+public abstract class RewardBundleEntry
+{
+    public abstract RewardBundleEntryType Type { get; }
+
+    public bool Unknown;
+    public int IconId;
+    public int IconTintId;
+    public int NameId;
+    public int Quantity = 1;
+    public int DefinitionId;
+    public int Tint;
+    public string Text = string.Empty;
+    public int Unknown2;
+    public bool Unknown3;
+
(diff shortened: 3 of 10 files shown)


src/Sanctuary.Packet/BaseRewardPacket/RewardBundlePacket.cs
yungcomputerchair: `BaseRewardPacket` is actually named `RewardBasePacket` in the client; I know the misname happened in the first PR but we should update it for accuracy

src/Sanctuary.Packet/BaseRewardPacket/RewardBundleEntry.cs
```diff
@@ -0,0 +1,39 @@
+using Sanctuary.Core.IO;
+using Sanctuary.Packet.Common;
+
+namespace Sanctuary.Packet;
+
+public abstract class RewardBundleEntry
```
yungcomputerchair: This is called `RewardBundleEntryBase` in the client

src/Sanctuary.Packet/BaseRewardPacket/RewardBundleEntry.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    public abstract RewardBundleEntryType Type { get; }
+
+    public bool Unknown;
```
yungcomputerchair: This field is `IsHidden`

src/Sanctuary.Packet.Common/RewardBundleEntryBase.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    public abstract RewardBundleEntryType Type { get; }
+
+    public bool Unknown;
+    public int IconId;
+    public int IconTintId;
+    public int NameId;
+    public int Quantity = 1;
+    public int DefinitionId;
+    public int Tint;
```
yungcomputerchair, line 15: nit: this is conventionally called `TintId`

src/Sanctuary.Packet/BaseRewardPacket/RewardBundleEntry.cs
```diff
@@ -0,0 +1,39 @@
...
+    public int IconId;
+    public int IconTintId;
+    public int NameId;
+    public int Quantity = 1;
+    public int DefinitionId;
+    public int Tint;
+    public string Text = string.Empty;
```
yungcomputerchair: What exactly does `Text` do? Can we give it a more specific name?

src/Sanctuary.Packet/BaseRewardPacket/RewardBundleEntry.cs
```diff
@@ -0,0 +1,39 @@
...
+    public int Unknown2;
```
yungcomputerchair: `Unknown2` is an ARGB color, so we can probably rename this to `TextColor` (or if you change `Text`, whatever that new name is + `Color`)

src/Sanctuary.Packet/BaseRewardPacket/RewardBundleEntry.cs
```diff
@@ -0,0 +1,39 @@
...
+    public bool Unknown3;
```
yungcomputerchair: `Unknown3` is `MembersOnly`

src/Sanctuary.Packet/BaseRewardPacket/RewardBundleEntryItem.cs
```diff
@@ -0,0 +1,17 @@
...
+
+public sealed class RewardBundleEntryItem : RewardBundleEntry
+{
+    public override RewardBundleEntryType Type => RewardBundleEntryType.Item;
+
+    public int ItemGuid;
+
+    protected override void SerializeData(PacketWriter writer)
+    {
+        // Success controls the client's subtype-data flag; it is not another wire field.
```
yungcomputerchair: wtf does this comment mean

also, I think a comment here acknowledging how `RewardBundleEntryItem` is serialized as just the item GUID even though it inherits from `RewardBundleEntryBase` would be pretty valuable (or at least a lot more valuable than whatever your AI left :danklol:)

yungcomputerchair (changes requested):
Just some minor field naming.

Also, do you know how `RewardBundleBase` (**not** `RewardBundleEntryBase`) is supposed to fit into all of this? Apparently it's used in the client's representation of `ClientCollection`; something along the lines of:

```
public class ClientCollection : ISerializableType
{
    public int Id;

    public int NameId;
    public int DescriptionId;

    /// <summary>
    /// Stored in client-side resource file <c>CollectionCategories.txt</c>
    /// </summary>
    public int CategoryId;

    public IconData Icon = new();

    /// <summary>
    /// 0 - ?
    /// 1 - ?
    /// </summary>
    public int State;

    public RewardBundleBase RewardBundle = new();

    public Dictionary<int, ClientCollectionEntry> Entries = [];

    public void Serialize(PacketWriter writer)
    {
        writer.Write(Id);

        writer.Write(NameId);
        writer.Write(DescriptionId);

        writer.Write(CategoryId);

        Icon.Serialize(writer);

        writer.Write(State);

        RewardBundle.Serialize(writer);

        writer.Write(Entries);
    }
}
```

src/Sanctuary.Packet.Common/RewardBundleEntryType.cs
yungcomputerchair, line 1: Not sure if we want to just include the entire enum since it's ripped from the client:
```
enum REWARDBUNDLE_TYPE
{
  REWARDBUNDLE_TYPE_ITEM = 0x1,
  REWARDBUNDLE_TYPE_EXPERIENCE = 0x3,
  REWARDBUNDLE_TYPE_QUESTADD = 0x6,
  REWARDBUNDLE_TYPE_PROFILEADD = 0x7,
  REWARDBUNDLE_TYPE_ABILITYLINE = 0x8,
  REWARDBUNDLE_TYPE_COLLECTIONADD = 0xA,
  REWARDBUNDLE_TYPE_COLLECTIONENTRYADD = 0xB,
  REWARDBUNDLE_TYPE_TOKEN = 0xC,
  REWARDBUNDLE_TYPE_PETTRICKEXPERIENCE = 0xD,
  REWARDBUNDLE_TYPE_RECIPE = 0xE,
  REWARDBUNDLE_TYPE_ZONEFLAG = 0xF,
  REWARDBUNDLE_TYPE_CHARACTERFLAG = 0x11,
  REWARDBUNDLE_TYPE_WHEELSPIN = 0x12,
  REWARDBUNDLE_TYPE_TROPHY = 0x13,
  REWARDBUNDLE_TYPE_CLIENTEXITURL = 0x14,
  REWARDBUNDLE_TYPE_FACTORYEXPERIENCE = 0x16,
  REWARDBUNDLE_TYPE_CHEST = 0x17,
};
```

Maybe unused enum variants cause warnings? If not I think we should.

yungcomputerchair:
I'm not sure why the build timed out, I've noticed that happening intermittently. I'll run it again for you

yungcomputerchair (approved):
Seems to check out and has been tested. This will be leveraged in `ClientCollection` in a follow-up PR.

PR #106, opened 2026-08-08
Fix mounts being stranded in the world when summoning while already riding

MyRealms:
## Problem

Summoning a mount while already riding one leaves the previous mount behind. It
stays standing in the world, visible to every nearby player, and never
disappears. A player can repeat this indefinitely and litter the zone.

The most convenient way to hit it is the store: buy a mount while riding and
press **Use Now**.

## Cause

`SpawnMount` assigns `Player.Mount` without checking whether the player is
already riding.

A mount registers itself in three places — the zone's npc map, the zone's entity
map, and its zone tile — but the only handle pointing back at it is
`Player.Mount`. Both cleanup paths go through that field:

```csharp
Player.Dismount()  ->  if (Mount is null) return; ... Mount.Dispose();
Player.Dispose()   ->  Mount?.Dispose();
```

Overwriting the field severs the only route to the previous mount. It remains
registered in all three collections and on every nearby client, with nothing
able to reach it again. It is not a GC leak — the zone still holds the object —
but the code responsible for removing it can no longer find it.

`PacketMountSpawnByItemDefinitionIdHandler` (the Use Now path) dispatches into
the same function, which is why both entry points are affected.

## Fix

Dismount first, at the top of `SpawnMount`, so both callers are covered.
`Player.Dismount()` already does the right thing: dismount response, stat reset,
graceful removal packet, dispose. The on-foot stats it restores are overwritten
by the new mount's a few lines further down.

## Testing

Reproduced with several players on a local server, then verified fixed: mounting
from the mount list and buying with Use Now both cleanly replace the current
mount, with nothing left behind. Repeated summons no longer accumulate.


GroaxyVRC:
This is exactly what I was going to do to fix but seems you got to it before I did. Not entirely sure why it was working without this before but seems something shifted and started producing the bug

StimpyDev:
The mount stat change isn't needed, only the dismount call is enough

src/Sanctuary.Gateway/Handlers/MountBasePacket/PacketMountSpawnHandler.cs
GroaxyVRC, line 109: Not entirely sure on this change? There's only 3 mounts outside of all the flying mounts that change "JumpHeight" and 1 of them isn't a flying mount. The other 2 is the Fire Hawk and S.M.A.C.M

GroaxyVRC: The Fire Hawk and SMACM have different values in all fields from the regular flying mounts

MyRealms:
Let me correct that.

yungcomputerchair (approved):

PR #107, opened 2026-08-10
Added new tints/tint groups to some items.

ImAlko:
- All max-level combat armor comes in coin store & station cash tints
- Bomber jacket, orange knee-high boots and gold-buckled knee-high boots are now associated with freestyle jobs
- cat and dog ears got additional tints

PR #108, opened 2026-08-10
Clean shaven facial card works now

ImAlko:
The clean shaven card doesn't have a style parameter (`Param2`), so it defaults to `0`. The mappings file for facial hair does not have an entry for `0`, so both the preview and use handlers for style cards was returning null, preventing the packet from being properly sent. Now, it'll catch the `0` and return an empty string. 

Side note: lots of duplicate logic in those two files. Not sure if there's a reason for that.

### DIFF
--- a/src/Resources/CoinStoreItems.json
+++ b/src/Resources/CoinStoreItems.json
@@ -294,7 +294,8 @@
   },
   {
     "Id": 208,
-    "CategoryId": 105
+    "CategoryId": 105,
+    "TintGroupId": 1
   },
   {
     "Id": 209,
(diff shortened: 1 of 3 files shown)


StimpyDev:
Nice catch. 

Close #107 and change the title for this pr to mention both changes. But that's up to you

yungcomputerchair:
Yeah I think your changes in #107 got picked up here. I'm with Stimpy, no harm in combining the PRs imo.

Edit: I still appreciate that you put up separate PRs; it makes it easy to revert the changes individually if something goes wrong. Keep that up. But I think we're safe for this one.

ImAlko:
> Nice catch. 
> 
> 
> 
> Close #107 and change the title for this pr to mention both changes. But that's up to you

Ah! Silly me I branched off of my coin shop branch :( Good catch. I can combine them though! Standby, I'll try to get some more items in the shop before a merge

yungcomputerchair (approved):

PR #109, opened 2026-08-10
Add quest definitions, JSON loading and the CharacterQuests table

Sulphural:
Data model and resource loading for quests, plus the persistence schema they will use. 

Quests are authored in Resources/Quests.json: text ids, giver/turn-in NPCs, rewards, and an ordered list of goals (talk to an NPC, reach a location, or collect pickups).

QuestDefinitionCollection indexes quests by giver and by every goal target, and assigns world guids to each Collect goal's pickups so an interaction can be traced back to the goal it credits.


yungcomputerchair:
This is a perfect first step. Thank you!! I'll put in a review soon.

src/Sanctuary.Database/Entities/DbCharacterQuest.cs
```diff
@@ -0,0 +1,21 @@
...
+public class DbCharacterQuest
+{
+    public int QuestId { get; set; }
+
+    public ulong CharacterId { get; set; }
+    public DbCharacter Character { get; set; } = null!;
+
+    public bool Completed { get; set; }
+
+    // How many of the quest's goals are done. Goals tick off in order, so 0 = on the first one.
```
yungcomputerchair: nit: I think you can delete these comments as the fields are self-explanatory. If you insist on keeping them, please at least make them proper doc comments.

src/Sanctuary.Database/Entities/DbCharacterQuest.cs
```diff
@@ -0,0 +1,21 @@
...
+    public int GoalProgress { get; set; }
+
+    // Pickups gathered so far, when the active goal is a Collect goal.
+    public int GoalCount { get; set; }
+
+    // The character's tracked quest, i.e. the breadcrumb / "Take Me There" target. At most one row
+    // per character is true.
+    public bool IsActive { get; set; }
```
yungcomputerchair: This is a real design flaw, and the comment even flags it. Having the active signal on the quest itself allows you to mathematically represent multiple active quests on the character at once, which should never be possible.

We should instead have a nullable ActiveQuestId column in the character table. That ensures that it's only possible to represent a single active quest at a time.

src/Sanctuary.Game/Resources/Definitions/QuestDefinition.cs
```diff
@@ -0,0 +1,96 @@
+using System.Collections.Generic;
+
+namespace Sanctuary.Game.Resources.Definitions;
+
+public sealed class QuestDefinition
+{
+    public int QuestId { get; set; }
+
+    // Only to keep Quests.json readable - the client resolves TitleId instead.
+    public string? Name { get; set; }
```
yungcomputerchair: Remove if unused in the C#; no need to deserialize cosmetic fields in the JSON

src/Sanctuary.Game/Resources/Definitions/QuestDefinition.cs
```diff
@@ -0,0 +1,96 @@
...
+    public int NotificationActive { get; set; } = 6;
+
+    // Hover cursor for the giver / turn-in NPC. 17 is the quest cursor (collection nodes use 18).
+    public byte NpcCursorId { get; set; } = 17;
+
+    // How close the player has to be for the client to accept a click on that NPC.
+    public int NpcInteractRange { get; set; } = 12;
+
+    // The authored goals, or the legacy single "talk to the target NPC" goal for quests written before
+    // Goals existed. Never empty for a well-formed quest.
```
yungcomputerchair: What does this comment mean by "before goals existed"? This is the first PR to ever introduce questing, so all quests should have well-defined goals. If this is a migration based off your local branch it should be removed since it's just noise that will never run on the main branch.

src/Sanctuary.Game/Resources/Definitions/QuestDefinition.cs
```diff
@@ -0,0 +1,96 @@
...
+                }
+            ];
+
+    // The turn-in bubble: the final goal's line, since a multi-goal quest can end back at the giver.
+    public int TurnInDialogueId
+    {
+        get
+        {
+            var goals = EffectiveGoals;
+            var last = goals[goals.Count - 1];
```
yungcomputerchair: I believe this will crash the server if this property is accessed when there are no goals. There should be some kind of guard here or alternative to unsafe indexing

src/Sanctuary.Game/Resources/Definitions/QuestGoal.cs
```diff
@@ -0,0 +1,105 @@
...
+namespace Sanctuary.Game.Resources.Definitions;
+
+public enum QuestGoalType
+{
+    TalkToNpc = 0,
+    ReachLocation = 1,
+    Collect = 2
+}
+
+// One turn of an NPC conversation. The player's reply is the caption on the dialog's response button.
```
yungcomputerchair: Remove or formalize as doc comment

src/Sanctuary.Game/Resources/Definitions/QuestGoal.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    // For Collect and counted talk goals. 0 = CollectSpawns.Count.
+    public int RequiredCount { get; set; }
+
+    public int CollectModelId { get; set; }
+    public int CollectNameId { get; set; }
+
+    // Hover cursor and click distance for the pickups, same wiring collection nodes use.
+    public byte CollectCursorId { get; set; } = 17;
+    public int CollectInteractRange { get; set; } = 12;
```
yungcomputerchair: Why are these both here and also in the quest definition? They should ideally only be in one place

src/Sanctuary.Game/Resources/QuestDefinitionCollection.cs
```diff
@@ -0,0 +1,186 @@
...
+
+using Microsoft.Extensions.Logging;
+
+using Sanctuary.Game.Resources.Definitions;
+
+namespace Sanctuary.Game.Resources;
+
+// One collectible pickup to spawn for a Collect goal. Guids are assigned at load time and map back to
+// (quest, goal) via Collectibles.
+public sealed class CollectibleSpawn
```
yungcomputerchair: @amuralle some of this looks like repeated work that the collections system already put in. can you take a look?

src/Sanctuary.Game/Resources/QuestDefinitionCollection.cs
```diff
@@ -0,0 +1,186 @@
...
+
+    public QuestDefinitionCollection(ILogger logger)
+    {
+        _logger = logger;
+    }
+
+    public bool TryGet(int questId, out QuestDefinition definition) => Quests.TryGetValue(questId, out definition!);
+
+    // The tightest interact range any quest using this NPC asks for, since one NPC often serves several.
+    // False when it isn't a quest NPC, leaving the entity's own default alone.
```
yungcomputerchair: make into doc comment

yungcomputerchair (changes requested):
I haven't looked into the logic too deeply just yet but I spotted a few issues that need to be addressed.

src/Sanctuary.Game/Resources/QuestDefinitionCollection.cs
Sulphural: Removed collections stuff from PR.. but the current system is global with a respawn timer. Definitely need to discuss more on this.

amuralle: Yeah I've been thinking about this- there's collectables that are user-specific used in quests for sure, I remember tons of them that only the quester could see. Will start looking through my own stuff and see if we can fit this in somehow.

src/Sanctuary.Game/Resources/Definitions/QuestGoal.cs
yungcomputerchair: The dedupe here went the wrong way. These stats are per-goal, not per-quest. So move them back to the QuestGoal and use them from there.

I'm actually not sure it makes sense for these interactable values to be in the quest domain at all; we should probably have an `IInteractable` interface that unifies this and the collection NPCs (@amuralle FYI)

src/Sanctuary.Game/Resources/QuestDefinitionCollection.cs
yungcomputerchair: I think we need to all have a talk about how this should be architected

src/Sanctuary.Game/Resources/Definitions/QuestGoal.cs
yungcomputerchair: (to be clear, you don't have to worry about an interface right now. Just move these to `QuestGoal` for now)

PR #110, opened 2026-08-12
Add Referees

cillow19:
### Summary
This PR adds the referee profile to users who are admins/moderators, and removes the profile from users who are not. The referee job notably provides the referee T-shirt as a default item. All referees will also have the pink display color **_in the perspective of the observer_** (more on this below). 

### Changes
- New Referee profile definition with default items.
- `ProfileHelper.cs` - added a special profile helper in preparation for enforcers, or other special jobs
- `PacketLoginHandler.cs` - on login, grants the Referee profile if user is mod/admin, and revokes it otherwise (covers demotion)
- `IsReferee `flag and active profile ID set in `Player.cs`
- Re-sends the `AddPc` packet when a referee switches profile, so nearby players' clients update immediately.

### Notes:
- Recently promoted/demoted users must log in again to gain/lose the referee profile.
- Unfortunately, I'm unable to find the Referee title or the Referee job profile name. Due to this, referees will not have the title and the referee job will be a second "adventurer" with the referee icon (thanks to ycc!). If anyone knows the name IDs, that would be amazing and I could add it either in this PR or a separate one.
- I wasn't able to fully replicate some behaviors on the admin side. For example, in the POV of the referee, the pink display name will only be present on the referee job, while in the POV of the observers, referees will have pink display names regardless of what job they are on. I'm pretty unsure how to proceed from here. I have tried some cheesy methods, but didn't want to impact the code too much for a small thing like name colors. **If anyone has suggestions on this in case I missed/misunderstood something, please do share!!!**

Let me know if you have any questions + concerns. All feedback is appreciated!

Issue: https://github.com/Open-Source-Free-Realms/Sanctuary/issues/64


GroaxyVRC:
There is no way as the client code has removed the pink name from showing on all classes, its only visible on other clients because of the isReferee on the AddPC code, I'm not too sure why the original developers changed this but its something editz looked into

cillow19:
> There is no way as the client code has removed the pink name from showing on all classes, its only visible on other clients because of the isReferee on the AddPC code, I'm not too sure why the original developers changed this but its something editz looked into

Gotcha, thanks for explaining. Think this will just be something we will have to deal with for now... I'll leave it up to the @yungcomputerchair and rest of staff on if we would like to continue pursuing enforcers/referees knowing this gap.

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -568,10 +568,11 @@ public PlayerUpdatePacketAddPc GetAddPcPacket()
 
             IsUnderage = Age < 18,
             IsMember = MembershipStatus != 0,
+            IsReferee = isReferee,
 
             // playerUpdatePacketAddPc.TemporaryAppearance = 277;
 
-            ActiveProfileId = ActiveProfileId,
+            ActiveProfileId = isReferee ? (int)SpecialProfileIds.Referee : ActiveProfileId,
```
yungcomputerchair: Won't this make all admins and mods appear to have the Referee profile always equipped?

src/Sanctuary.Gateway/GatewayConnection.cs
```diff
@@ -612,7 +612,6 @@ public void SendSelfToClient()
         Player.Collections = _resourceManager.Collections.CreateClientCollections(Player.Guid, ownedItemDefinitionIds);
 
         var packetSendSelfToClient = new PacketSendSelfToClient();
-
```
yungcomputerchair, line 615: nit: random whitespace change

src/Sanctuary.Gateway/Handlers/BaseCommandPacket/CommandPacketSetProfileHandler.cs
```diff
@@ -65,6 +67,9 @@ public static bool HandlePacket(GatewayConnection connection, ReadOnlySpan<byte>
 
         connection.Player.SendTunneledToVisible(playerUpdatePacketEquippedItemsChange);
 
+        if (isReferee)
+            connection.Player.SendTunneledToVisible(connection.Player.GetAddPcPacket());
```
yungcomputerchair: How come this is needed?

src/Sanctuary.Game/Helpers/ProfileHelper.cs
```diff
@@ -0,0 +1,108 @@
...
+    {
+        foreach (var defaultItemId in profileData.DefaultItems)
+        {
+            if (!resourceManager.ClientItemDefinitions.TryGetValue(defaultItemId, out var defaultClientItemDefinition))
+                continue;
+
+            if (defaultClientItemDefinition.GenderUsage != 0 && defaultClientItemDefinition.GenderUsage != character.Gender)
+                continue;
+
+            var dbItem = character.Items.SingleOrDefault(x => x.Definition == defaultItemId);
```
yungcomputerchair: `SingleOrDefault` will throw if there are multiple results. You probably want `FirstOrDefault`

src/Sanctuary.Game/Helpers/ProfileHelper.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        foreach (var item in character.Items)
+        {
+            if (!existingItemIds.Contains(item.Id))
+                dbContext.Entry(item).State = EntityState.Added;
+        }
+
+        dbContext.Entry(newProfile).State = EntityState.Added;
+
+        dbContext.SaveChanges();
```
yungcomputerchair: I think you want to check the result of this and have some way to indicate a failure to the caller (and I also think it might throw; I would look at the other calls in the codebase for guidance).

src/Sanctuary.Game/Helpers/ProfileHelper.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        var profileToRemove = character.Profiles.FirstOrDefault(p => p.Id == id);
+        if (profileToRemove != null)
+        {
+            character.Profiles.Remove(profileToRemove);
+        }
+    }
+    // void AddTitle();
+    // void RemoveTitle();
```
yungcomputerchair: Is this future work?

src/Sanctuary.Game/Helpers/ProfileHelper.cs
```diff
@@ -0,0 +1,108 @@
...
+        if (!resourceManager.Profiles.TryGetValue(id, out var profileData))
+        {
+            logger.LogWarning("Profile with ID {profileId} does not exist in the resource manager.", id);
+            return;
+        }
+
+        // Check if the character already has the profile
+        if (character.Profiles.Any(p => p.Id == id))
+        {
+            logger.LogInformation("Character {characterId} already has profile {profileId}.", character.Id, id);
```
yungcomputerchair: This should definitely be debug level. It'll trigger a lot, right?

yungcomputerchair (changes requested):
Looks mostly good, just needs a few tweaks.

src/Sanctuary.Game/Entities/Player.cs
GroaxyVRC: There's no need for that as the isReferee already does that, what i mean is client sided only the referee job gives a pink name

cillow19: this has been removed and reverted back to just assign ActiveProfileId to ActiveProfileId

src/Sanctuary.Game/Helpers/ProfileHelper.cs
cillow19: I left it commented in case someone knew the title Id for referees, but since this is not likely, I have removed the comments

src/Sanctuary.Gateway/Handlers/BaseCommandPacket/CommandPacketSetProfileHandler.cs
cillow19: ah my bad, think i was playing around with something and forgot to get rid of it. it has been removed now.

src/Sanctuary.Game/Helpers/ProfileHelper.cs
cillow19: thanks for pointing this out - i have added a try catch similar to other calls in the codebase... let me know if that's sufficient!!!

src/Sanctuary.Game/Helpers/ProfileHelper.cs
```diff
@@ -0,0 +1,119 @@
...
+        character.Profiles.Add(newProfile);
+        return true;
+    }
+
+    public static void RemoveSpecialProfile(DbCharacter character, DatabaseContext dbContext, SpecialProfileIds profileId)
+    {
+        int id = (int)profileId;
+
+        dbContext.Profiles.Where(p => p.CharacterId == character.Id && p.Id == id).ExecuteDelete();
+        dbContext.SaveChanges();
```
yungcomputerchair: Need the same try/catch and result check here as above

src/Sanctuary.Gateway/Handlers/BaseCommandPacket/CommandPacketSetProfileHandler.cs
```diff
@@ -65,6 +67,7 @@ public static bool HandlePacket(GatewayConnection connection, ReadOnlySpan<byte>
 
         connection.Player.SendTunneledToVisible(playerUpdatePacketEquippedItemsChange);
 
+
```
yungcomputerchair: nit: whitespace

yungcomputerchair (changes requested):
Got one more DB call to guard but besides that looks good 👍

src/Sanctuary.Game/Helpers/ProfileHelper.cs
cillow19: i removed SaveChanges and added a try catch ExecuteDelete instead. let me know if that works!

yungcomputerchair (approved):
LGTM. I would test it again to make sure it still works after these iterations if you haven't already. If it looks good I'll merge. Let me know

yungcomputerchair:
(actually you have perms, so just squash and merge at your discretion)

cillow19:
> (actually you have perms, so just squash and merge at your discretion)

just tested it on my test server - looks good so far. merging! 
thanks for the helpful feedback!!!

PR #112, opened 2026-08-15
BUG FIX: Sell items in the shop

raisingkaines:
This change updates the sell-to-shop handler so one request can sell an entire inventory stack, not just a fixed quantity.
What it does (CoinStoreBuyFromClientRequestPacketHandler.cs):
Reads the requested packet.Quantity from the client.
If Quantity <= 0, it treats that as “sell all” and sets sellQuantity to clientItem.Count.
Otherwise it uses the requested quantity as before.
It then validates stock, removes the items, credits coins for sellQuantity, updates DB/client counts, and records the transaction with sellQuantity.
In short: it now supports a “sell everything in this stack” path while preserving existing exact-quantity sells.
How it differs from upstream repo (upstream/main):
Upstream only allowed the explicit quantity path:it checked if (clientItem.Count < packet.Quantity)
decremented by packet.Quantity
paid ResellValue * packet.Quantity
recorded transaction Quantity = packet.Quantity

### DIFF
--- a/src/Sanctuary.Gateway/Handlers/BaseCoinStorePacket/CoinStoreBuyFromClientRequestPacketHandler.cs
+++ b/src/Sanctuary.Gateway/Handlers/BaseCoinStorePacket/CoinStoreBuyFromClientRequestPacketHandler.cs
@@ -58,7 +58,9 @@ public static class CoinStoreBuyFromClientRequestPacketHandler
             return true;
         }
 
-        if (clientItem.Count < packet.Quantity)
+        var sellQuantity = packet.Quantity <= 0 ? clientItem.Count : packet.Quantity;
+
+        if (clientItem.Count < sellQuantity)
         {
             coinStoreTransactionCompletePacket.Result = 4;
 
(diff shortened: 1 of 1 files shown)


src/Sanctuary.Gateway/Handlers/BaseCoinStorePacket/CoinStoreBuyFromClientRequestPacketHandler.cs
```diff
@@ -163,12 +165,12 @@ public static bool HandlePacket(GatewayConnection connection, ReadOnlySpan<byte>
...
 
         connection.SendTunneled(coinStoreTransactionCompletePacket);
 
         connection.Player.CoinStoreTransactions.Add(coinStoreTransactionCompletePacket.TransactionRecord);
 
         return true;
     }
-}
\ No newline at end of file
+}
```
yungcomputerchair, line 176: nit: whitespace change

yungcomputerchair (approved):
LGTM

yungcomputerchair:
For patch notes: is there a "sell all" button in the client that sends a non-positive quantity value? I.e. does this benefit players or is this a dev scenario?

raisingkaines:
This is a player-facing change. It tells the server to treat Quantity <= 0 as “sell the whole stack” in one request, so it’s not a dev-only path.

yungcomputerchair:
> This is a player-facing change. It tells the server to treat Quantity <= 0 as “sell the whole stack” in one request, so it’s not a dev-only path.

Right, but what does the UX look like from the player side? They don't have to specify 0 or a negative value in the quantity box, do they?

raisingkaines:
I do appear to be wrong as 0 would not be the default as it always starts at 1 when selling.

StimpyDev:
Selling already works, the reason why selling isn't working is because every item has a resell value of 0. So all u have to do is change the resell value in the json file. This has been tested.

yungcomputerchair:
Since this is not the actual behavior exhibited by the client (there's no "sell all" button that sends a non-positive value in the packet) we probably don't want this behavior serverside since it's not accommodating anything.

PR #113, opened 2026-08-16
Finalizing prototype of drop table and rewards manager.

ImAlko:
This PR adds a general weighted drop table to use for collections and rewards (so far). Additionally, this expands upon the work done on rewards by creating a rewards manager that handles dishing out various types of rewards to players. T

### DIFF
--- a/src/Sanctuary.Game/Resources/RewardTableDefinitionCollection.cs
+++ b/src/Sanctuary.Game/Resources/RewardTableDefinitionCollection.cs
@@ -49,7 +49,8 @@ public sealed class RewardTableDefinitionCollection : ObservableConcurrentDictio
             {
                 if (string.IsNullOrWhiteSpace(entry.Key) || entry.DropTable.Count == 0 ||
                     entry.DropTable.Any(drop => drop.Weight <= 0) ||
-                    entry.DropTable.OfType<CurrencyRewardDropDefinition>().Any(drop => !Enum.IsDefined(drop.CurrencyType)))
+                    entry.DropTable.OfType<CurrencyRewardDropDefinition>().Any(drop => !Enum.IsDefined(drop.CurrencyType)) ||
+                    entry.DropTable.OfType<ItemRewardDropDefinition>().Any(drop => drop.Quantity <= 0))
                 {
                     _logger.LogError("Invalid reward table in \"{file}\".", filePath);
                     return false;
--- a/src/Sanctuary.Game/Resources/RewardTableDefinitionCollection.cs
+++ b/src/Sanctuary.Game/Resources/RewardTableDefinitionCollection.cs
@@ -1,0 +1,84 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text.Json;
+
+using Microsoft.Extensions.Logging;
+
+using Sanctuary.Core.Collections;
+using Sanctuary.Game.Resources.Definitions.Rewards;
+
+namespace Sanctuary.Game.Resources;
+
+public sealed class RewardTableDefinitionCollection : ObservableConcurrentDictionary<string, RewardTableDefinition>
+{
+    private readonly ILogger _logger;
+    private readonly object _writeLock = new();
+
+    public RewardTableDefinitionCollection(ILogger logger)
+    {
+        _logger = logger;
+    }
+
+    public bool Load(string filePath)
+    {
+        if (!File.Exists(filePath))
+        {
+            _logger.LogError("Failed to find file \"{file}\"", filePath);
+            return false;
+        }
+
+        try
+        {
+            using var stream = new FileStream(filePath, FileMode.Open, FileAccess.Read, FileShare.Read);
+            var entries = JsonSerializer.Deserialize<List<RewardTableDefinition>>(stream, new JsonSerializerOptions
+            {
+                PropertyNameCaseInsensitive = true
+            });
+
+            if (entries is null || entries.Count == 0)
+            {
+                _logger.LogError("No reward tables found in \"{file}\".", filePath);
+                return false;
+            }
+
+            var loaded = new Dictionary<string, RewardTableDefinition>(StringComparer.OrdinalIgnoreCase);
+
+            foreach (var entry in entries)
+            {
+                if (string.IsNullOrWhiteSpace(entry.Key) || entry.DropTable.Count == 0 ||
+                    entry.DropTable.Any(drop => drop.Weight <= 0) ||
+                    entry.DropTable.OfType<CurrencyRewardDropDefinition>().Any(drop => !Enum.IsDefined(drop.CurrencyType)))
+                {
+                    _logger.LogError("Invalid reward table in \"{file}\".", filePath);
+                    return false;
+                }
+
+                entry.Key = entry.Key.Trim().ToLowerInvariant();
+
+                if (!loaded.TryAdd(entry.Key, entry))
+                {
+                    _logger.LogError("Duplicate reward table {key} in \"{file}\".", entry.Key, filePath);
+                    return false;
+                }
+            }
+
+            lock (_writeLock)
+            {
+                foreach (var entry in loaded)
+                    this[entry.Key] = entry.Value;
+
+                foreach (var key in Keys.Where(key => !loaded.ContainsKey(key)).ToArray())
+                    Remove(key);
+            }
+
+            return true;
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Failed to parse file \"{file}\".", filePath);
+            return false;
+        }
+    }
+}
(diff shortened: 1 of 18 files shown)


amuralle:
Got some visibility on this now. This looks super compatible with what I’m doing with rewards for collections too.

The only thing I’d flag is that collections currently have their own item grant + reward toast flow, so after this lands I can wire collections into `IRewardManager` instead of letting those paths drift separately.

Future-direction wise, I think this PR is probably setting up `IRewardManager` as the shared gameplay reward path, which feels right. For collections, I’d expect the collection system to keep handling collection-specific progress/notifications, while `IRewardManager` handles the actual item/currency grant and item reward toast. Later we can extend it with more reward entry types as we validate them, especially for non-item collections like exploration coins.

I'll add a couple comments on file specific stuff too but nice 👍

src/Sanctuary.Game/RewardManager.cs
```diff
@@ -0,0 +1,189 @@
...
+        return drop switch
+        {
+            ItemRewardDropDefinition item => TryGrantItem(player, item.ItemDefinitionId,
+                item.TintTable?.SelectRandom().TintId ?? 0, sourceGuid),
+            CurrencyRewardDropDefinition currency => TryGrantCurrency(player, currency.CurrencyType, currency.Amount),
+            _ => false
+        };
+    }
+
+    public bool TryGrantItem(Player player, int itemDefinitionId, int tint, ulong sourceGuid = 0)
```
amuralle: Tiny future API thought: this currently grants quantity `1`. That’s fine for the current table shape, but if this is becoming the general grant path we may want `quantity` on item reward drops / `TryGrantItem`.

amuralle: Related broader thought on this same method: this looks like the right place for collections to eventually call into too. It overlaps with the item grant + inventory update + reward toast flow that collections and shop purchase paths also do. I don’t think this PR has to resolve that fully, but I’d like to treat `IRewardManager` as the future shared path so those flows don’t drift. Part of the reason I'd want this to land first.

src/Sanctuary.Game/RewardManager.cs
```diff
@@ -0,0 +1,189 @@
...
+            return false;
+
+        player.Coins = dbCharacter.Coins;
+
+        player.SendTunneled(new ClientUpdatePacketCoinCount { Coins = player.Coins });
+
+        return true;
+    }
+
+    private static void SendRewardToast(Player player, ClientItem clientItem, ClientItemDefinition itemDefinition, ulong sourceGuid)
```
amuralle: Bigger design question for the team: what client-facing feedback should `RewardManager` own?

My instinct is that it should own the universal/generic reward feedback, like the standard reward-bundle item toast, because that seems tied to the act of granting the reward itself. More specific UI, like collection started / collection entry added / collection complete for example, probably belongs in the respective systems that triggered the reward. So for questing- quests would also want to own their own specific notifications- that's the theoretical line we should be aware of- but I like this. 

I don’t think we need a separate UI abstraction yet unless reward feedback starts diverging heavily. If most rewards share the same generic toast path, keeping that packet send close to the grant logic seems reasonable.

Going to comment some additions worth adding on this point

src/Sanctuary.Game/RewardManager.cs
```diff
@@ -0,0 +1,189 @@
...
+    {
+        // TODO: Is there a coin/station cash toast? How about for other types of rewards..?
```
amuralle: Coins definitely have a toast- it's something we'll want to add, Station Cash I don't think so since that ever was a reward for anything in the vanilla game. 

the coin toast did an effect on the player model that showered them in coins in addition to an item toast and a little coin shower audio clip, I'll bet we can track both the effects down in the effects section and maybe expand the item grant to do coins too WITH that effect on top? 

On the subject of SC though-  maybe we will need to come up with something to reward station cash and then a notif for that given we aren't doing a purchasing route.... 

Other than items and coins though, I'm not aware of anything else that would be needed right now.

src/Sanctuary.Game/Resources/RewardTableDefinitionCollection.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                _logger.LogError("No reward tables found in \"{file}\".", filePath);
+                return false;
+            }
+
+            var loaded = new Dictionary<string, RewardTableDefinition>(StringComparer.OrdinalIgnoreCase);
+
+            foreach (var entry in entries)
+            {
+                if (string.IsNullOrWhiteSpace(entry.Key) || entry.DropTable.Count == 0 ||
```
amuralle: Personal taste here, but I’d probably lean toward some #failfastmaxxing on reward config by forcing `WeightedDropTable` / nested tint table construction during resource load. Since the tables are config-backed, I’d rather find out at startup that something is busted than only discover it when a player hits the first roll/grant.

src/Sanctuary.Game/RewardManager.cs
```diff
@@ -0,0 +1,189 @@
...
+        if (dbCharacter is null)
+            return false;
+
+        var dbItem = dbCharacter.Items.SingleOrDefault(item => item.Definition == itemDefinitionId && item.Tint == tint);
+
+        if (dbItem is null)
+        {
+            dbItem = new DbItem
+            {
+                Id = dbCharacter.Items.Select(item => item.Id).DefaultIfEmpty(0).Max() + 1,
```
amuralle: Tiny future-proofing/edge-casing thought: if this becomes the shared reward path, should item id assignment eventually go through some kind of locked helper? Mostly thinking about the freak case where two grants both do `Max() + 1` at the same time and pick the same id. Unclear if this would/could actually happen tho.

src/Sanctuary.Game/RewardManager.cs
ImAlko: > Tiny future API thought: this currently grants quantity `1`. That’s fine for the current table shape, but if this is becoming the general grant path we may want `quantity` on item reward drops / `TryGrantItem`.

Added quantities to items. This is a great catch!

ImAlko: > Related broader thought on this same method: this looks like the right place for collections to eventually call into too. It overlaps with the item grant + inventory update + reward toast flow that collections and shop purchase paths also do. I don’t think this PR has to resolve that fully, but I’d like to treat `IRewardManager` as the future shared path so those flows don’t drift. Part of the reason I'd want this to land first.

Yeah I actually made a comment within `GrantItem` because I expected that we have multiple similar logic paths for item granting. I'm going to keep it outside of this PR right now (unless someone really wants me to handle it). I think it's important that we unify that as well :)

src/Sanctuary.Game/RewardManager.cs
```diff
@@ -0,0 +1,190 @@
...
+                item.TintTable?.SelectRandom().TintId ?? 0, item.Quantity, sourceGuid),
+            CurrencyRewardDropDefinition currency => TryGrantCurrency(player, currency.CurrencyType, currency.Amount),
+            _ => false
+        };
+    }
+
+    public bool TryGrantItem(Player player, int itemDefinitionId, int tint, int quantity, ulong sourceGuid = 0)
+    {
+        // NOTE: I think similar logic lives in the shop + collections stuff. We should eventually extract
+        // and generalize that to avoid repeats.
```
yungcomputerchair: we should make an issue for this and make this a TODO so we can track it

src/Sanctuary.Game/RewardManager.cs
```diff
@@ -0,0 +1,190 @@
...
+
+    public bool TryGrantItem(Player player, int itemDefinitionId, int tint, int quantity, ulong sourceGuid = 0)
+    {
+        // NOTE: I think similar logic lives in the shop + collections stuff. We should eventually extract
+        // and generalize that to avoid repeats. 
+        if (!_resourceManager.ClientItemDefinitions.TryGetValue(itemDefinitionId, out var itemDefinition))
+        {
+            _logger.LogError("Cannot grant unknown item definition {itemDefinitionId}.", itemDefinitionId);
+            return false;
+        }
```
yungcomputerchair, line 66: Do you really need the load-time check if you have this?

yungcomputerchair (approved):
Looks great to me! Well done on the design

src/Sanctuary.Game/RewardManager.cs
ImAlko: I'd say its borderline redundant. _Maybe_ worth keeping in as the drop table load-time check can help catch issues with the `json`. Let me know if you'd prefer I remove it.

src/Sanctuary.Game/RewardManager.cs
ImAlko: Done

PR #111, opened 2026-08-15
Add housing editor, directory, and zoning fixes

raisingkaines:
Housing Editor, Directory, Ownership, and Zoning Update
This update adds the complete server-side housing system currently developed for Sanctuary. It includes character-based home ownership, the native decoration interface, persistent fixture editing, housing directories and ratings, housing-item purchasing, dedicated housing zones, and several important zoning and stability fixes.
Home ownership and decoration UI
Housing ownership is stored per character rather than per account. Each character maintains its own collection of homes, fixtures, customizations, ratings, and inventory.
Two characters on the same account can own the same type of home independently, while a single character cannot purchase the same home definition twice.
Home owners receive the packets required to open and use the game client’s native decoration UI, including:
Housing edit grants
Housing fixture inventory
Current fixture and landmark counts
Furniture score
House information and customization state
Persisted fixture definitions and assets
Edit-mode state changes
Visitors can enter and view an available home, but decoration operations verify ownership again on the server. Visitors cannot place, move, recolor, save, customize, or remove another character’s fixtures.
Complete housing editor support
The update implements the full housing fixture workflow:
Begin fixture placement
Create and manage placement previews
Place fixtures
Move and rotate fixtures
Change fixture scale
Save fixture transformations
Pick up individual fixtures
Pick up all fixtures
Apply customizations by fixture group and type
Remove fixture customizations
Persist fixture tint, texture, position, rotation, and scale
Restore fixtures after zoning or reconnecting
Refresh the owner’s fixture inventory after purchases or editing
Maintain stable links between housing fixtures and their world actors
Persisted fixture actors are reconstructed after the client finishes loading. Their server-side and client-side identities remain stable, preventing missing models, broken selection, stale collision, and fixtures that cannot be moved after re-entering a home.
Fixture collision and editor selection are handled separately. Placement previews remain non-collidable until promoted to saved fixtures, while saved building pieces restore their proper collision and editor raycast behavior.
Interactive housing fixtures
Non-musical interactive housing objects retain their runtime behavior, including:
Teleporters and launch pads
Moving elevators and platforms
Train sets
Gumball machines
Fireworks
Party pools and synchronized water companion models
Fixture interactions are limited to the correct player and active house instance. Stale or forged actor identifiers cannot be used to interact with objects belonging to another home.
Housing directory and ratings
The update includes the native housing directory protocol and database support for:
Published homes
Featured homes
Friends’ homes
Search and candidate results
House descriptions and keyword lists
House ratings
Vote totals
[...]

PR #116, opened 2026-08-20
First pass at generic zoning

ImAlko:
TODO: Fill this out

TODO: Talk about limitations, future work, etc


ImAlko:
@yungcomputerchair let me know if you think this should be broken up. I'm not very happy with the size and amount of files I touched

ImAlko:
Consider removing any of the housing DB stuff and replace with a comment or something. It's _necessary_, but isn't fully in scope with what the PR is trying to do (would love external opinions on this)

yungcomputerchair:
Marked as a blocker for #111

src/Sanctuary.Game/ZoneManager.cs
```diff
@@ -76,23 +87,147 @@ public bool TryGetPlayer(string name, [MaybeNullWhen(false)] out Player player)
...
+
+        // Two callers can race to create the first instance for the same (zoneDefinitionId, ownerId)
+        // pair - e.g. two party members both zoning into a not-yet-created dungeon at once. Whichever
+        // TryAdd loses just discards its own zone (never started, so this is cheap) and hands back
+        // whichever instance actually won.
+        if (!_zones.TryAdd(key, newZone))
+        {
+            newZone.Dispose();
+            return _zones.TryGetValue(key, out zone);
+        }
```
yungcomputerchair: I'm still a little bit nervous because of TOC/TOU here. I think instead of `TryGet` at the start of the function and `TryAdd` here, it would be better to leverage the APIs that `ConcurrentDictionary` has for this type of thing that already handles the concurrency for you (for example, [GetOrAdd()](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=net-10.0#system-collections-concurrent-concurrentdictionary-2-getoradd(-0-1)))

src/Sanctuary.Game/ZoneManager.cs
```diff
@@ -76,23 +87,147 @@ public bool TryGetPlayer(string name, [MaybeNullWhen(false)] out Player player)
...
+        zone = new HousingZone(housingZoneDefinition, _serviceProvider)
+        {
+            Id = _uniqueId++,
+            OwnerId = ownerId
+        };
+
+        return true;
+    }
+
+    public bool TryGrantHouse(int zoneDefinitionId, ulong ownerId)
```
yungcomputerchair: This function feels kind of out of place here since it's mostly DB stuff. Maybe hoist the DB logic somewhere else and call it from here? But then you have a very tiny function. Idk.

yungcomputerchair (changes requested):
I generally like the architecture, and the inheritance model makes sense to me. Nice work. I know the runtime data model for housing and combat is basically empty right now, but I don't think that's a code smell; those will gain more and more fields as their respective features are implemented.

I definitely think you need to leverage `ConcurrentDictionary`'s built-in functionality a little more; it's safer and introduces less overhead than doing our own check, and will keep your code simpler. For now, that's the only blocker I see in the actual instancing logic.

I would strongly recommend stripping out the DB stuff. I know you said it's necessary, but this PR really does feel like two disjoint sets of changes (housing ownership and zone instancing). I would keep this PR scoped to the latter and allow housing ownership to be tied in later. Right now, bare housing zones with no DB persistence is totally fine IMO.

src/Sanctuary.Game/ZoneManager.cs
yungcomputerchair: and if you use those you probably don't need to check disposal state which is a lot cleaner :)

ImAlko:
Running the new `!zone ls` command

In `FabledRealms`
image

In `hsg_emptylot_seaside_beach_01`
image

src/Sanctuary.Game/ZoneManager.cs
ImAlko: It's removed and the responsibility of housing folks to implement!

src/Sanctuary.Game/ZoneManager.cs
ImAlko: Attempted a cleanup. Should hopefully be a little easier to read now

yungcomputerchair:
Looks way better, thanks. I'll give it another review soon.

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -209,20 +212,25 @@ private void UpdateZoneTile()
...
     }
 
-    public void TeleportToZone(IZone zone, Vector4 position, Quaternion rotation)
+    public bool TeleportToZone(IZone zone, Vector4 position, Quaternion rotation)
     {
         if (Zone == zone)
-            return;
+            return true;
+
+        if (zone.IsDisposed)
```
yungcomputerchair: we should never be holding a reference to a disposed zone, let alone passing it into `TeleportToZone()`. we should see if we can get rid of this

src/Sanctuary.Game/ZoneManager.cs
```diff
@@ -15,13 +21,15 @@ public class ZoneManager : IZoneManager
     private readonly ILogger _logger;
     private readonly IResourceManager _resourceManager;
     private readonly IServiceProvider _serviceProvider;
-
     private static int _uniqueId = 1;
 
-    private readonly ConcurrentDictionary<int, IZone> _zones = new();
+    private readonly ConcurrentDictionary<(int, ulong?), IZone> _zones = new();
+
+    public int StartingZoneDefinitionId => 1;
```
yungcomputerchair, line 31: nit: make constant instead of property

src/Sanctuary.Game/ZoneManager.cs
```diff
@@ -76,23 +84,103 @@ public bool TryGetPlayer(string name, [MaybeNullWhen(false)] out Player player)
...
-        zone = new StartingZone(startingZoneDefinition, _serviceProvider)
+        zone = new WorldZone(worldZoneDefinition, _serviceProvider)
         {
             Id = _uniqueId++
         };
 
-        zone.OnStart();
+        zone.Start();
+
+        return _zones.TryAdd((zone.DefinitionId, null), zone);
```
yungcomputerchair: need to flip these so `TryAdd()` is called first, return if it fails, and THEN start the zone.

src/Sanctuary.Game/ZoneManager.cs
```diff
@@ -76,23 +84,103 @@ public bool TryGetPlayer(string name, [MaybeNullWhen(false)] out Player player)
...
+
+    public bool TryGetOrCreateZoneInstance(int zoneDefinitionId, ulong? ownerId, [MaybeNullWhen(false)] out IZone zone)
+    {
+        var key = (zoneDefinitionId, ownerId);
+
+        var storedZone = _zones.GetOrAdd(key, _ => CreateZoneInstance(zoneDefinitionId, ownerId)!);
+
+        if (storedZone is null)
+        {
+            _zones.TryRemove(key, out _);
```
yungcomputerchair: don't need this, if `GetOrAdd` fails it won't be in the map.

yungcomputerchair: actually you don't need this block at all. `GetOrAdd()` never returns null. It'll either return the newly added zone or the existing instance.

src/Sanctuary.Game/Zones/IZone.cs
```diff
@@ -13,9 +13,18 @@ namespace Sanctuary.Game.Zones;
...
+    Vector4 SpawnPosition { get; }
+    Quaternion SpawnRotation { get; }
+
+    bool IsDisposed { get; }
+    bool TryMarkDisposedIfEmpty();
+
     #region Events
 
-    void OnStart();
+    void Start();
```
yungcomputerchair: we should just keep this as `OnStart()` and move all the thread spinup logic in `BaseZone` into its `OnStart()`. child classes can override and call the base class `OnStart()` first if they really want to, and non-BaseZone impls can figure their own stuff out.

src/Sanctuary.Game/ZoneManager.cs
```diff
@@ -15,13 +21,15 @@ public class ZoneManager : IZoneManager
     private readonly ILogger _logger;
     private readonly IResourceManager _resourceManager;
     private readonly IServiceProvider _serviceProvider;
-
     private static int _uniqueId = 1;
```
yungcomputerchair, line 25: we need to make this atomic

src/Sanctuary.Game/ZoneManager.cs
```diff
@@ -76,23 +84,103 @@ public bool TryGetPlayer(string name, [MaybeNullWhen(false)] out Player player)
...
+            },
+            _ => null
+        };
+    }
+
+    public void RemoveZoneInstance(IZone zone)
+    {
+        var key = (zone.DefinitionId, zone.OwnerId);
+
+        ((ICollection<KeyValuePair<(int, ulong?), IZone>>)_zones).Remove(new(key, zone));
```
yungcomputerchair: what is this cast lol

src/Sanctuary.Game/ZoneManager.cs
```diff
@@ -76,23 +84,103 @@ public bool TryGetPlayer(string name, [MaybeNullWhen(false)] out Player player)
...
+
+    private bool TryGetHousingZone(HousingZoneDefinition housingZoneDefinition, ulong ownerId,
+        [MaybeNullWhen(false)] out HousingZone zone)
+    {
+        // NOTE: As of PR-116, this MIGHT NOT be hooked up properly. It will be the responsibility
+        // of anyone working on housing to properly match conventions/make adjustments. Once that is 
+        // done, remove this note.
+        //
+        // The commented code will be left as a convenience. You will need to give database access
+        // to the class for it to work.
```
yungcomputerchair: nit: you can really shorten this comment, the commented out code below makes it obvious what needs to be done here.
```suggestion
        // TODO: fetch from DB
```

yungcomputerchair (changes requested):
from on-call code review

src/Sanctuary.Game/ZoneManager.cs
ImAlko: So sometimes `CreateZoneInstance` will return `null`, e.g., we request a house for a given player, but that player doesn't own that house. I _guess_ that can be gated higher up in the logic, but this is why this block exists.

src/Sanctuary.Game/ZoneManager.cs
ImAlko: I think we run into some issues (since other things depend on the interface level) when making this constant. Okay to leave it like this (should _function_ the same way)?

src/Sanctuary.Game/Zones/IZone.cs
ImAlko: Okay, I made this change as we discussed!

src/Sanctuary.Game/ZoneManager.cs
ImAlko: Likely my AI slop haha. Replaced it with something that _should_ work equivalently

src/Sanctuary.Game/ZoneManager.cs
ImAlko: Should be fixed.

src/Sanctuary.Game/ZoneManager.cs
ImAlko: Done

src/Sanctuary.Game/ZoneManager.cs
ImAlko: Should be done!

src/Sanctuary.Game/Entities/Player.cs
ImAlko: So there is a potential race and that's why this exists. The zone's `OnSecond` loop checks for empty players and disposes itself if that's the case (this will eventually have something more robust like a grace period). So I think there's a rare chance where we grab a zone that has just disposed itself.

ImAlko:
My AI caught a couple of other things I should clean up. Will get to those later tonight

src/Sanctuary.Game/Entities/Player.cs
yungcomputerchair: I think it's dangerous to have zones be in charge of disposing themselves and managing their lifecycles. It should be the responsibility of the zone manager to evict empty zones: when an entity is transferred out, it can filter out zones with zero entity count out of the concurrent dictionary. It also serves as the perfect point of synchronization if we really do need to put a lock in place.

src/Sanctuary.Game/ZoneManager.cs
yungcomputerchair: wondering if we can put `readonly` on the int so the compiler can stop us from accidentally accessing it directly

src/Sanctuary.Game/ZoneManager.cs
yungcomputerchair: yeah that's fine

src/Sanctuary.Game/ZoneManager.cs
yungcomputerchair: hmmmm. My concern is that this is starting to get really messy. There are a ton of `Try` calls and at each one, we need to worry about failure and figuring out how to clean up. I strongly think this should be rearchitected to be less fallible; in this case, check if the player owns the house before even thinking about zone creation.

ImAlko: Okay, yeah I do agree its a bit messy, but I'm struggling to find something that _isn't_ a little messy. I will take a pass at a slightly different architecture tomorrow when my mind is fresh. I'm sure there is a way to clean this up, but due to the edge cases for each type of zone it may not be the _cleanest_.

src/Sanctuary.Game/ZoneManager.cs
ImAlko: We can't do this because `readonly` isn't compatable with passing this value by reference to `.Increment`. We can probably ensure this protection if you want, but it may be over-engineered if I do so (would have to look into it).

src/Sanctuary.Game/Entities/Player.cs
ImAlko: You're absolutely right, thanks! I've went ahead and made the `ZoneManager` responsible for this and by extension whatever removes a player calls this function to check.

src/Sanctuary.Game/ZoneManager.cs
yungcomputerchair: ew. Thanks .NET. Ok, this is fine for now then.

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -209,20 +212,31 @@ private void UpdateZoneTile()
...
+            if (!_zoneManager.TryGetOrCreateZoneInstance(zone.DefinitionId, zone.OwnerId, out var freshZone))
+                return false;
+
+            zone = freshZone;
+        }
+
+        // NOTE: There may be a tiny period of time in which
+        // the player exists in the dictionary of BOTH zones,
+        // new and old.
+        if (!zone.TryAddPlayer(this))
```
yungcomputerchair: We absolutely cannot have this. There are multiple spots in the codebase that call into the ZoneManager to find a player, and they assume that there's one and only one instance of the player.

src/Sanctuary.Game/Entities/Player.cs
yungcomputerchair: This looks better, but exposing `IsDisposed` on the zone still feels like an overreach. What exactly is this checking for? Can we instead add an API to the ZoneManager that takes in the `IZone` and tells us if it's being tracked or not? That will always be accurate since the ZoneManager is the source of truth now.

Better yet, all of this zoning code should probably live in the zone manager itself instead of in the entity classes...

yungcomputerchair (changes requested):
I think we're getting closer to a good system here, but there's still some things bothering me.

src/Sanctuary.Game/Entities/Player.cs
ImAlko: Good thing I added this note! Will look into this

src/Sanctuary.Game/ZoneManager.cs
```diff
@@ -76,23 +89,143 @@ public bool TryGetPlayer(string name, [MaybeNullWhen(false)] out Player player)
...
+
+        // using var dbContext = _dbContextFactory.CreateDbContext();
+
+        // var dbHouse = dbContext.Houses.SingleOrDefault(house =>
+        //     house.CharacterId == ownerId && house.ZoneDefinitionId == zoneDefinitionId);
+
+        // if (dbHouse is null)
+        //     return false;
+
+        return true;
```
strophiccat: this is an issue. anyone who joins a housing zone or instance becomes the owner of it. would it not make sense to throw a NotImplementedException() or at least return false in the interim until this function is implemented?

src/Sanctuary.Game/Zones/BaseZone.cs
strophiccat, line 1: architecture here looks overall good, but two things i'd like to mention about maintainability

i dont think OnClientIsReady should be in the base. every future zone (dungeon, house, minigame) inherits the full overworld send and has to suppress it. WorldZone already exists as the renamed StartingZone, so moving the block there instead should be a small change and keeps the base thin.

there's also +-1000 lines of hardcoded journal/welcome/stat data in here. i believe that should be migrated to a json file and read through IResourceManager like the other definition tables.

src/Sanctuary.Game/Zones/IZone.cs
```diff
@@ -13,11 +13,18 @@ namespace Sanctuary.Game.Zones;
...
+
+    Vector4 SpawnPosition { get; }
+    Quaternion SpawnRotation { get; }
+
     #region Events
 
     void OnStart();
     void OnClientIsReady(Player entity);
     void OnClientFinishedLoading(Player entity);
+    void Dispose();
```
strophiccat: nit: just extend `IDisposable` instead of declaring manually. grants `using`support and analyzer warnings for missed dispoal

src/Sanctuary.Game/Zones/BaseZone.cs
```diff
@@ -880,17 +2181,26 @@ private async Task UpdateEverySecondAsync()
...
+        }
     }
 
     public void Dispose()
     {
+        // NOTE: We currently do not have a grace-period or timeout for this. This
+        // function WILL get called if no players exist in a zone (except for FabledRealms)
+        // which is the 'Starting Zone'. If things slow down, then this might get called
+        // before a player makes it in.
+        _zoneManager.RemoveZoneInstance(this);
```
strophiccat: `Dispose()` takes the zone out of `_zones` before it cancels the tick tasks, so for a moment it's gone from the manager but still running. it's also public on `IZone` now, so someone could call it directly and skip the `_playerTransitionLock` that `EvictIfEmpty` uses.

might be cleaner to have one teardown path behind the lock that stops the ticking first, then clears everything, then removes it.

src/Sanctuary.Game/ZoneManager.cs
ImAlko: This returned `true` so I could test moving to houses. I think for the purposes of actually submitting this PR, you're right. It's safer to return `false`. _Or_ the _right_ option is probably a `NotImplementedException` as you said. Will make this change!

src/Sanctuary.Game/Zones/BaseZone.cs
ImAlko: Yeah, good catch. Should be fixed now.

Regarding the public `Dispose`... I made a few functions that are _technically_ behind locks in the manager public so that I could simplify and call them directly in the manager. I'm basically creating a contract with the other devs that they will use the zone manager as intended, rather than calling these functions willy-nilly. 

If you have a better solution, let me know.

src/Sanctuary.Game/Zones/IZone.cs
ImAlko: Done

src/Sanctuary.Game/Zones/BaseZone.cs
ImAlko: I don't think we can move _everything_ out of `BaseZone`, but I do think you are right. Some of it can live in in `WorldZone`. I've went ahead and made those changes.

FYI: I did not hardcode those values. They were just moved it from the `StartingZone`, which predates me. I was going to leave it as is and we can just have a follow-up PR that cleans that stuff up. Would we be okay with this?

ImAlko:
I noticed an issue...

The welcome UI thing shows up whenever you teleport into a zone (specifically a `WorldZone` because that's where `SendWelcomeInfo` lives now). Yes, that is literally how it's coded right now. 

How do we want to solve this? I can add a flag onto `Player`, then check it in `WorldZone` before `SendWelcomeInfo` is called. Flag is triggered on login.

However, we should _also_ probably put `SendWelcomeInfo` back into `BaseZone` and send the welcome UI no matter what zone you're in when you log in. What do people think?

image

src/Sanctuary.Game/Zones/IZone.cs
```diff
@@ -26,8 +33,12 @@ public interface IZone : IScriptableZone
     IEnumerable<Npc> Npcs { get; }
     IEnumerable<Player> Players { get; }
 
+    bool IsEmpty { get; }
+
     float TickDeltaSeconds { get; }
 
+
+
     bool TryGetNpc(ulong guid, [MaybeNullWhen(false)] out Npc npc);
```
yungcomputerchair, line 42: nit: blank lines

src/Sanctuary.Game/ZoneManager.cs
```diff
@@ -76,23 +89,147 @@ public bool TryGetPlayer(string name, [MaybeNullWhen(false)] out Player player)
...
+        lock (_playerTransitionLock)
+        {
+
+            if (zone.IsEmpty)
+                zone.Dispose();
+        }
+    }
+
+
+    private bool PlayerCanAccessZone(BaseZoneDefinition definition, ulong? ownerId) => definition switch
```
yungcomputerchair, line 203: this and `PlayerOwnsHouse` should be `static`

src/Sanctuary.Game/Entities/Mount.cs
```diff
@@ -21,8 +21,13 @@ public Mount(IZone zone, Player rider, MountDefinition definition) : base(zone)
         Definition = definition;
     }
 
-    public override void TeleportToZone(IZone zone, Vector4 position, Quaternion rotation)
+    public override bool TeleportToZone(IZone zone, Vector4 position, Quaternion rotation)
     {
+        Zone.TryRemoveNpc(Guid);
+
+        if (!zone.TryAddMount(this))
+            return false;
```
yungcomputerchair, line 29: shouldn't this call happen after the `Visible = false;` line? otherwise all the logic for broadcasting the mount's removal gets skipped. I understand that having the entity in limbo is not great but having them be visually orphaned is pretty bad too

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -249,21 +254,23 @@ public void TeleportToZone(IZone zone, Vector4 position, Quaternion rotation)
             Name = Zone.Name,
             Position = position,
             Rotation = rotation,
-            Sky = "sky_deep_mines.xml",
+            Sky = null,
```
yungcomputerchair, line 257: Is this change intentional?

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -210,29 +213,31 @@ private void UpdateZoneTile()
...
-            return;
+        if (Zone == destinationZone)
+            return true;
+
+
+        var oldZone = Zone;
+
+        if (!_zoneManager.TryMovePlayerToZone(destinationZone.DefinitionId, destinationZone.OwnerId, this, out var zone))
+            return false;
+        _zoneManager.EvictIfEmpty(oldZone);
```
yungcomputerchair, line 226: shouldn't the zone manager just make this check whenever it moves an entity? not sure this should be called externally

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -210,29 +213,31 @@ private void UpdateZoneTile()
...
     {
-        if (Zone == zone)
-            return;
+        if (Zone == destinationZone)
+            return true;
+
+
+        var oldZone = Zone;
+
+        if (!_zoneManager.TryMovePlayerToZone(destinationZone.DefinitionId, destinationZone.OwnerId, this, out var zone))
```
yungcomputerchair, line 224: same issue as mounts here; we should unwind visibility *before* attempting to move the player. especially if the old zone is disposed

src/Sanctuary.Game/Zones/BaseZone.cs
```diff
@@ -902,5 +1368,8 @@ public void Dispose()
         _players.Clear();
```
yungcomputerchair, line 1368: Can disposal happen while the zone's update loops are running? that concerns me a bit, especially since `_tiles` is not a concurrent dictionary.

I'm also a little worried about cleanup; what happens to the NPCs left in the zone? I guess it's fine that we don't send out visibility packets, but I want to triple-check that no strong references are leaked when the zone is disposed. Some of those tendrils might exist in the form of script contexts as well... those NPCs need to all be disposed of properly and I don't think they are right now

yungcomputerchair (changes requested):

src/Sanctuary.Game/Zones/BaseZone.cs
ImAlko: We should now await both loops before we start the rest of the cleanup. `Dispose` is no longer under a lock to avoid potential bottlenecks that might come with this fix. We also make calls to `Npc.Dipose`.

src/Sanctuary.Game/Entities/Player.cs
ImAlko: Player was a bit more complicated, so things had to be massaged a bit so that the proper order of checking if a new zone could actually exist -> removing the player -> adding the player to the new zone would actually work without them disappearing on other clients.

Note: I'm not sure I like the solution, but it's what I came up with.

src/Sanctuary.Game/Entities/Mount.cs
ImAlko: This should be fixed.

src/Sanctuary.Game/Entities/Player.cs
ImAlko: Yes, because the other zones ended up with this sky. Changed it so the sky's definition actually gets propagated here

src/Sanctuary.Game/ZoneManager.cs
ImAlko: Fixed

src/Sanctuary.Game/Entities/Player.cs
ImAlko: We now call `EvictIfEmpty` _outside_ the lock when we try to move the player in the `ZoneManager`!

src/Sanctuary.Game/ZoneManager.cs
```diff
@@ -76,23 +89,161 @@ public bool TryGetPlayer(string name, [MaybeNullWhen(false)] out Player player)
...
+        if (!_zones.TryAdd((zone.DefinitionId, null), zone))
+            return false;
+
         zone.OnStart();
 
-        return _zones.TryAdd(zone.Id, zone);
+        return true;
+    }
+
+    public bool TryMovePlayerToZone(int zoneDefinitionId, ulong? ownerId, Player player, out IZone zone)
```
amuralle: I know follow-up item 3 already calls out the broader thread-safety problem around player position, tile, and zone. I’m wondering whether this specific transition boundary needs to be handled in this PR, though. We publish the player in the destination zone under _playerTransitionLock, but their Zone, ZoneTile, and position aren’t updated until this returns to Player.TeleportToZone(). That leaves the new zone briefly owning a player who still points at the old zone, and oldZone is also captured before the lock. Is the intent to defer this under item 3, or should the manager own the complete state swap and rollback now?

src/Sanctuary.Game/Zones/BaseZone.cs
```diff
@@ -885,27 +1353,36 @@ private async Task UpdateEverySecondAsync()
...
-        return _nextNpcGuid++;
+            return _nextNpcGuid++;
+        }
     }
 
     public void Dispose()
     {
         _cancellationTokenSource.Cancel();
 
+        Task.WaitAll(_updateEveryTickTask, _updateEverySecondTask);
```
amuralle: I see why we wait for both loops here after the earlier disposal discussion, but I don’t think cancellation currently wakes either one. 

If I'm understanding this right- the cancellation token is only checked after the timer produces its next tick. It is not passed to WaitForNextTickAsync().
That produces this sequence:
1. The final player leaves a disposable zone.
2. EvictIfEmpty() removes the zone and calls zone.Dispose().
3. Dispose() cancels the token.
4. The one-second loop is still asleep inside WaitForNextTickAsync().
5. Task.WaitAll() blocks until that timer naturally wakes.
6. Only then does the loop see the cancelled token and exit.

so then- that could stall the gateway thread for close to a second. Could we pass the token into the timer waits and handle normal cancellation, or dispose the timers before waiting?

src/Sanctuary.Game/Zones/BaseZone.cs
```diff
@@ -236,14 +668,35 @@ public bool TryCreateNpc([MaybeNullWhen(false)] out Npc npc)
         return TryCreateNpc(null, out npc);
     }
 
+    private bool TryRegisterEntity<TEntity>(ConcurrentDictionary<ulong, TEntity> collection, TEntity entity)
```
amuralle: hey, good helper - looks like TryAddPlayer and TryAddMount aren’t using it right now

src/Sanctuary.Game/ZoneManager.cs
```diff
@@ -76,23 +89,161 @@ public bool TryGetPlayer(string name, [MaybeNullWhen(false)] out Player player)
...
+        if (!PlayerCanAccessZone(zoneDefinition, ownerId))
+        {
+            zone = null;
+            return false;
+        }
+
+        var key = (zoneDefinitionId, ownerId);
+
+        zone = _zones.GetOrAdd(key, _ => CreateZoneInstance(zoneDefinition, ownerId));
+        zone.OnStart();
```
amuralle: may just be my doom-minded brain here, but since the zone is already in _zones here, what happens if OnStart() throws? 

BaseZone sets _started before firing the start script/activating pools, so it looks like we’d keep a partially initialized zone that skips startup on the next attempt. Should we remove/dispose it here on failure and only consider it started after initialization succeeds?

only really will balloon if this happened across a bunch of unique owner/instance keys, but I’ve seen weirder server stuff happen.

amuralle (changes requested):
Looks pretty good- I know ycc and fellow goats of reviewing did a great job catching most of it. I have a few ultra-safe things to add- but they seem pretty minor: 

- The transition lock moves the player between zone dictionaries, but the player’s actual Zone, ZoneTile, and position are updated afterward in Player.TeleportToZone(). I’d either have the manager own that complete transaction or explicitly work out how this fits into follow-up item 3.
- Disposal waits for the update loops, but the timer waits don’t currently observe the cancellation token, so evicting a zone could stall the gateway thread until the one-second timer wakes.
- TryAddPlayer() and TryAddMount() look like they missed the new TryRegisterEntity() rollback helper.
- If OnStart() throws after the zone enters _zones, we appear to retain a partially initialized instance that will skip startup on subsequent attempts.
-
None of those make me question the overall direction or suggest another big rearchitecture. They mostly look like targeted hardening around the boundaries this PR is introducing. Once those are worked through, I feel pretty good about this as the base for the services that need instanced zones.

Nice work, dude! Let's knock out these small things and I'd be happy to see it on the server.

src/Sanctuary.Game/Zones/BaseZone.cs
ImAlko: Done! Great catch, you are right

src/Sanctuary.Game/ZoneManager.cs
ImAlko: This wasn't something I was going to handle in the follow up. It's real and I fixed it. Thanks!

src/Sanctuary.Game/Zones/BaseZone.cs
ImAlko: Oops, I never hooked it up. Fixed!

src/Sanctuary.Game/ZoneManager.cs
ImAlko: There's now a `Try`/`Catch` for this!

amuralle (approved):

PR #117, opened 2026-08-20
Fix: Disconnect idle clients from client metrics

raisingkaines:
760e4936d8d18f3b076237f3ed8c54a2
It's basically just a repair to the clients lacking AFK Idle timeout

### DIFF
--- a/src/Sanctuary.Gateway/Handlers/PacketClientMetricsHandler.cs
+++ b/src/Sanctuary.Gateway/Handlers/PacketClientMetricsHandler.cs
@@ -29,6 +29,13 @@ public static class PacketClientMetricsHandler
 
         _logger.LogTrace("Received {name} packet. ( {packet} )", nameof(PacketClientMetrics), packet);
 
+        if (packet.Metrics.Idle == 1)
+        {
+            _logger.LogInformation("{connection} disconnected for inactivity. ( Guid: {guid}, IdleTime: {idleTime} )", connection, connection.Player?.Guid, packet.Metrics.IdleTimeout);
+
+            connection.Disconnect();
+        }
+
         return true;
     }
 }


yungcomputerchair:
What are the exact circumstances needed for the client to set the idle flag (i.e. how long without activity, and what constitutes activity)? We should have a solid, documented understanding of the client behavior here since this can interrupt the player experience by killing connections.

Also, we probably want this to be a release-mode-only mechanism.

raisingkaines:
i checked the client behavior in the decompile. the idle flag gets set after 600 seconds, or 10 minutes, without an input event reaching the client’s input queue. keyboard presses and releases, mouse movement, mouse clicks, double-clicks, and the mouse wheel all reset the timer. simply focusing the window, mouse enter and leave events or using the device outside the game do not necessarily count as activity. the client sends metrics about every 15 seconds, so the disconnect should happen around 10 minutes to 10 minutes and 15 seconds after the last recognized input. the 24-hour value is a separate maximum connection time and is unrelated to afk behavior. this submission is mostly to solve the issue of AFK chat drama lurkers in game as I've been approached about a few times.

src/Sanctuary.Gateway/Handlers/PacketClientMetricsHandler.cs
```diff
@@ -29,6 +29,13 @@ public static bool HandlePacket(GatewayConnection connection, Span<byte> data)
...
+        if (packet.Metrics.Idle == 1)
+        {
+            _logger.LogInformation("{connection} disconnected for inactivity. ( Guid: {guid}, IdleTime: {idleTime} )", connection, connection.Player?.Guid, packet.Metrics.IdleTimeout);
+
+            connection.Disconnect();
+        }
+
         return true;
     }
-}
```
yungcomputerchair, line 34: nit: whitespace change

yungcomputerchair (approved):
Can you include a screenshot of the client UX at disconnect in the PR description please?

raisingkaines:
done

raisingkaines:
removed empty space

PR #119, opened 2026-08-23
First prototype for NPC behaviors.

ImAlko:
Please read...

The purpose of this PR is to prototype an interface between the `lua` layer and `C#` layer for NPC behavior. I'm not fully happy with it, but in case it shows potential, I want to have this up. I _do not_ expect this to go through, but I'm just listing some considerations I've been making about NPCs and how they might behave.

# Actions
Why? Some NPC behaviors are instant (i.e., saying something or dropping a reward), while others may need to run on the tick (or per second) loop (e.g., movement, waiting). An action tries to unify this. Each action would have the following:

1) `void OnStart`: called when the action is first started (great for setting things up once)
2) `bool OnTick`: called every tick, returns `true` when considered complete

This way, something like `lua` simply needs to construct a behavior (or set of behaviors) and pass it through. Everything will be stepped `OnTick`, so nothing should be blocking the process.

## Managing actions
The `ActionManager` keeps a dictionary of string keys to actions and updates every tick. When an action completes, it is removed from the dictionary (and thus, doesn't run). If one needed to interrupt an action (e.g., the NPC is moving towards the player, but something triggers asking it to retreat), one can simply replace the value in the dictionary for a given key. 

## Action types
1) `InstantAction`: performs some function and immediately returns `true`
2) `TimeoutAction`: performs an action with a timeout that will return `true`
3) `WaitAction`: non-blocking action used for waiting some amount of time
4) `SequentialAction`: runs things one-by-one, then returns `true` when complete
5) `ParallelAction`: runs actions in parallel (i.e., each `OnTick` step is done on the same tick, this is _not_ true parallelism), then returns `true` when all are done

A `DelegateAction` can be used to define more custom behavior. For example, a `MoveToAction` can be built this way for NPCs to move.

## Future action types
In the interest of saving time, I have yet to implement other potentially (?) useful types of actions:
1) `RepeatAction`: Repeat a given action multiple times (or infinitely until interrupted) 
2) `WaitUntilAction`: Wait until a given condition is met, then perform the action
3) `DoUntilAction`: Perform an action until a given condition is met.
 
## References
1) https://docs.wpilib.org/en/stable/docs/software/commandbased/commands.html
2) https://rr.brott.dev/docs/v1-0/actions/

### DIFF
--- a/src/Sanctuary.Core/Actions/DelegateAction.cs
+++ b/src/Sanctuary.Core/Actions/DelegateAction.cs
@@ -1,0 +1,18 @@
+using System;
+
+namespace Sanctuary.Core.Actions;
+
+public sealed class DelegateAction : IAction
+{
+    private readonly Action _onStart;
+    private readonly Func<bool> _onTick;
+
+    public DelegateAction(Action? onStart = null, Func<bool>? onTick = null)
+    {
+        _onStart = onStart ?? (() => { });
+        _onTick = onTick ?? (() => true);
+    }
+
+    public void OnStart() => _onStart();
+    public bool OnTick() => _onTick();
+}
--- a/src/Sanctuary.Core/Actions/IAction.cs
+++ b/src/Sanctuary.Core/Actions/IAction.cs
@@ -1,0 +1,9 @@
+namespace Sanctuary.Core.Actions;
+
+public interface IAction
+{
+    void OnStart() { }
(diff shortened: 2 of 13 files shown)


yungcomputerchair:
talked on call about this, some good ideas but too much for now. we'll close and come back later as the need arises. tyty

PR #120, opened 2026-08-29
Exp granting & level-ups

ImAlko:
This PR adds progression to characters. Characters can now be given experience (stars) and they will see their level bar update as exp is given. Moreover, their level will increase if enough experience received. Currently, there is **one single level curve** for all jobs. But a per-job level curve can easily be massaged into the architecture if need be.

### Client-side changes (important)
As of this PR, the client has assets that behave strangely, namely the animations (and sounds) that correspond to a given level up. Job profile icons change at levels 5, 10, 15 and 20, matching the game on Sunset. However, special animations (and sounds) with these icons play on levels 4, 8, 12 and 16. This is _not_ something that should be fixed server-side. The client-side icons looking correct is an indication that the packet _seems_ correct.

Upon further research, it looks like `notifications.gfx` specifically hardcodes the animations to the "wrong" levels. Decompiling `ScriptsBase.bin` (client-side) also shows these "incorrect" mappings (this file is responsible for the scripts that trigger the _sounds_). To fix this, one would have to override `notifications.gfx` by using an _older_ version. I have found that one from the 2011 and 2012 clients works. Moreover, one would have to decompile `SciptsBase.bin`, adjust the mappings and re-compile to fix the sounds.


src/Sanctuary.Game/RewardManager.cs
```diff
@@ -163,6 +163,96 @@ public bool TryGrantCurrency(Player player, CurrencyType currencyType, int amoun
         return true;
     }
 
+    public bool TryGrantExperience(Player player, int profileId, int amount)
```
ImAlko: NOTE: I actually think a lot of this logic should potentially live in `Player.cs`. I'm finding it hard where to figure out where to draw the line between reward manager action and player reaction (right now all logic that happens during and after receiving a reward happens in this file, but maybe it shouldn't)

src/Sanctuary.Login/Handlers/CharacterCreateRequestHandler.cs
```diff
@@ -369,7 +369,7 @@ private static void UnlockAllProfiles(DbCharacter dbCharacter, DbItem chestItem,
             var dbProfile = new DbProfile
             {
                 Id = profileData.Id,
-                Level = 20
+                Level = 1
```
ImAlko, line 372: Should probably hardcode this back to 20, shouldn't I?

src/Sanctuary.Game/RewardManager.cs
```diff
@@ -163,6 +163,96 @@ public bool TryGrantCurrency(Player player, CurrencyType currencyType, int amoun
...
+        while (dbProfile.Level < maxLevel &&
+            _resourceManager.RankLevels.TryGetValue(dbProfile.Level, out var rankLevel) &&
+            dbProfile.LevelXP >= rankLevel.StarsToNextLevel)
+        {
+            dbProfile.LevelXP -= rankLevel.StarsToNextLevel;
+            dbProfile.Level++;
+        }
+    }
+
+    // Sent on every grant, level-up or not - refreshes the client's profile datasource
```
ImAlko: Remove AI comment

PR #121, opened 2026-09-01
Add combat job toolbars with starter weapon ability mappings

CarterW24:
Builds the ability bar from JSON on finish-loading, equip/unequip, and profile switch. Each combat job kit maps weapons to a basic and special ability; this ships the six starter weapons. Non-combat profiles get an empty set so the bar clears on job switch (verified live). Also fixes `Ability.Unknown8` -> `float Range` - reverse-engineered from the client, which reads the field as a float and uses it as the auto-target radius.

Ability execution is intentionally out of scope - next slice.


ImAlko:
Would you be able to post a video or something showing the job/weapon switches change the abilities in the bar? That'd be cool verification evidence!

src/Sanctuary.Game/Resources/AbilityDefinitionCollection.cs
```diff
@@ -0,0 +1,96 @@
...
+{
+    public static readonly IReadOnlySet<string> EffectTypes = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
+    {
+        "SweepDamage",
+        "AoeDamage",
+        "SingleTargetDamage",
+        "AoeDamageHeal",
+        "Summon",
+        "Buff",
+    };
```
cillow19: do you think an enum would be good here to avoid future typos/easier reference? i see that these strings are used in other definitions too...

CarterW24:
Removed 'CombatManager' and instead made Player own everything stateful.

'Update CombatAbilities.json' was to fix the unicode - \u0027 - in the json and replace them with " ' "

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -67,6 +68,97 @@ public sealed class Player : ClientPcData, IEntity
     public Vector4 StartingZonePosition { get; set; }
     public Quaternion StartingZoneRotation { get; set; }
 
+    #region Combat
+
+    public int GetEquippedWeaponDefinitionId()
+    {
+        const int PrimaryWeaponSlot = 7;
```
ImAlko: Magic number in the middle of a function.

I'd move this outside

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -67,6 +68,97 @@ public sealed class Player : ClientPcData, IEntity
...
+            SendTunneled(new AbilityPacketSetDefinition { ProfileId = ActiveProfileId });
+            return false;
+        }
+
+        var weaponDefinitionId = GetEquippedWeaponDefinitionId();
+        var (basic, special) = ResolveWeaponAbilities(kit, weaponDefinitionId);
+
+        var weaponNameId = 0;
+        if (_resourceManager.ClientItemDefinitions.TryGetValue(weaponDefinitionId, out var weaponDefinition))
+            weaponNameId = weaponDefinition.NameId;
```
yungcomputerchair: How does the client behave if this fails and remains zero?

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -67,6 +68,97 @@ public sealed class Player : ClientPcData, IEntity
...
+            DescriptionId = ability.DescriptionId,
+            IconId = ability.IconId,
+            ManaCost = ability.EnergyCost
+        });
+    }
+
+    private static Ability CreateToolbarSlot(int abilityDefinitionId, int iconId, int nameId, int manaCost) => new()
+    {
+        Type = 3,
+        Unknown2 = abilityDefinitionId,
```
yungcomputerchair: If this is known now, we should rename `Unknown2` to `AbilityDefinitionId`

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -67,6 +68,97 @@ public sealed class Player : ClientPcData, IEntity
     public Vector4 StartingZonePosition { get; set; }
     public Quaternion StartingZoneRotation { get; set; }
 
+    #region Combat
```
yungcomputerchair: nit: this region should go under the constructor. in general, try to keep the constructor above all functions in the file

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -181,6 +273,19 @@ public void UpdateEverySecond()
...
+        Energy = energy;
+
+        SendTunneled(new ClientUpdatePacketMana
+        {
+            CurrentMana = Energy,
+            MaxMana = maxEnergy
+        });
+    }
+
     public void UpdatePosition(Vector4 position, Quaternion rotation, bool updateZoneArea = true)
```
yungcomputerchair, line 185: not a fan of having `maxEnergy` as a parameter here, since a misinformed caller can override the max energy value from the kit on a whim. Can we instead have a private `MaxEnergy` property that is set by `SendToolbar()` and then just reference that here? I think that'll also let you collapse this logic into a real setter for the Energy property

src/Sanctuary.Game/IResourceManager.cs
```diff
@@ -1,4 +1,4 @@
-﻿using Sanctuary.Game.Resources;
+using Sanctuary.Game.Resources;
```
yungcomputerchair: nit: whitespace change

yungcomputerchair (changes requested):
Looks mostly good, only thing I'll block on is the energy comment

src/Sanctuary.Game/Entities/Player.cs
yungcomputerchair: tbh it would be nice if all new regions went near the bottom :)

src/Sanctuary.Game/Entities/Player.cs
yungcomputerchair: this will also let us reference the player's max energy later (which we will definitely need to do for stuff like regen)

yungcomputerchair: this will also let you cap `Energy` to `MaxEnergy` in the setter, instead of having to do the max logic inside `SendToolbar`!

yungcomputerchair (approved):

src/Sanctuary.Game/Entities/Player.cs
yungcomputerchair: resolved offline

PR #125, opened 2026-09-12
Fix food effects looping forever and set consumable cooldowns to 30 minutes

Sulphural:
Fixes a few consumable issues from #27.

- Food effects were sent with PlayCompositeEffect, which can't be stopped, and almost all of them are looping effects, so they never went away. They're now attached as effect tags and removed after 30 minutes. Eating another food replaces the current effect instead of stacking a second one.
- Steaming Pretzels used PFX_steam_white_rising, a positional one-shot that was sent without a position, so nothing showed up. Switched it to PFX_steam_white_rising_skel.
- Food and transformation cooldowns are now 30 minutes.
- Scaredy Cake sent every effect in a scare with Clear set, so each one wiped the one before it and only the last one played. The birds scare ended up as just a crow sound with nothing on screen. Only the first effect clears now.
- Cakes no longer repeat the outcome they just played. Scaredy Cake won't do the same scare or transformation twice in a row, and the Boss Birthday Cake won't give the same boss twice in a row.
- Added the Boss Birthday Cake. It places the boss birthday cake model in the world, idles and plays its ambient animation every so often, and clicking it turns you into Stone Heart, the Abominable Snowman or the Pumpkin Prince. It spawns with a white smoke and confetti burst. The one-shot animation timing is set per cake in Consumables.json.
- Both cakes now share a per-cake interact cooldown (InteractCooldownMs, 5 seconds), so the boss cake can no longer be spam-clicked. This replaces ScareCooldownMs.
- Cakes no longer play the smoke poof when their time runs out. They just disappear.
- The Realms Roll Boombox now turns players dancing to it into Rick and uses the rroll dance instead of looping freestyle_01. They change back when they walk away or the boombox ends, and anyone already transformed by something else keeps their current look.
- Updated the boombox list. The Rainbow Randomizer Boombox now plays one of its six colored music effects at random each time it is placed.
- Food effects now match what each item description says (hearts for Scrumptious Soup, ghosts for Deep Mines Delight, blue flames for Firefrog Legs and so on). Several were using one-shot effects or effects that had nothing to do with the food.
- Removed the boss cake setup that was attached to the Bat Cupcake. That item is a single-use Halloween cupcake, not the boss cake.


src/Sanctuary.Game/Resources/Definitions/BoomboxDefinition.cs
```diff
@@ -5,5 +5,7 @@ public class BoomboxDefinition
     public int ItemId { get; set; }
     public int ModelId { get; set; }
     public int EffectId { get; set; }
+    public int[] RandomEffectIds { get; set; } = [];
```
yungcomputerchair: Does it make sense to combine these into one `EffectIds` array and then just have an array of length 1 for the boomboxes with only one effect?

Now that I'm looking back at the JSON, all the boomboxes have an `EffectId` of 0. Is that even used then?

src/Sanctuary.Gateway/Helpers/Abilities/CakeAbility.cs
```diff
@@ -101,14 +114,79 @@ private void SpawnCakeNpc(Player player, CakeItemDefinition cakeDefinition)
             };
         }
 
-        BroadcastSpawn(player, cakeNpc, spawnPosition, cakeDefinition.SpawnPoofEffectId);
+        var spawnRecipients = BroadcastSpawn(player, cakeNpc, spawnPosition, cakeDefinition.SpawnEffectIds[0]);
```
yungcomputerchair: This will throw if `SpawnEffectIds` is empty. Should put a fallback value here.

src/Sanctuary.Gateway/Helpers/Abilities/CakeAbility.cs
```diff
@@ -54,40 +54,53 @@ private void SpawnCakeNpc(Player player, CakeItemDefinition cakeDefinition)
...
+                    for (var i = 0; i < scareEffects.Length; i++)
                     {
                         player.SendTunneledToVisible(new PlayerUpdatePacketPlayCompositeEffect
                         {
                             Guid = cakeNpc.Guid,
-                            CompositeEffectId = effectId,
+                            CompositeEffectId = scareEffects[i],
                             Position = cakeNpc.Position,
-                            Clear = true
+                            Clear = i == 0
```
yungcomputerchair, line 105: Can you explain which bug this `i == 0` clear fixes?

src/Sanctuary.Gateway/Helpers/Abilities/FoodEffectAbility.cs
```diff
@@ -36,10 +37,47 @@ public override bool HandleAbility(Player player, AbilityPacketClientRequestStar
...
+        if (delayMs > 0)
+            player.SendTunneledToVisibleDelayed(addAura, delayMs, true);
+        else
+            player.SendTunneledToVisible(addAura, true);
+
+        player.SendTunneledToVisibleDelayed(new PlayerUpdatePacketRemoveEffectTagCompositeEffect
+        {
+            Guid = player.Guid,
+            TagId = tagId
+        }, delayMs + FoodEffectDurationMs, true);
```
yungcomputerchair, line 81: This is fine for now, but just want to bring this up: in Free Realms, did food effects have any logical impact on the player?
This delayed packet framework is great for purely visual effects, but if it's anything more, this will need to be rewritten in the near future, since all this does is schedule a packet send.

yungcomputerchair (changes requested):
Looks mostly good, my only blocking concern is the possible throw on `SpawnEffectIds[0]`.

src/Sanctuary.Gateway/Helpers/Abilities/CakeAbility.cs
Sulphural: PlayCompositeEffect wipes every composite effect on the entity before playing the new one. The scare loop sent every effect in a group with Clear set, so each one erased the one before it and only the last effect was ever visible. The birds scare, for example, ended up as just the crow sound with nothing on screen. Clearing only on the first effect still removes whatever the previous scare left behind, and lets the rest of the group play together.

src/Sanctuary.Gateway/Helpers/Abilities/FoodEffectAbility.cs
Sulphural: As far as I can tell they were purely cosmetic. The item descriptions only mention visuals.

yungcomputerchair (approved):
Looks good! Thanks!!

PR #126, opened 2026-09-13
Show transformations and food effects on the buff bar

Sulphural:
Transformations and food effects now show up on the buff bar like they did in Free Realms: the change form icon, the item name, and a "Time left" countdown. Clicking the X on the buff ends it early.

- Added `ClientUpdatePacketAddEffectTag` (38/16) and `ClientUpdatePacketRemoveEffectTag` (38/17). The add packet is the tag id followed by a length-prefixed blob holding the `EffectTag` fields plus the icon id and name id, which is what the client reads.
- Added `PlayerUpdatePacketRequestStripEffect` (35/33) and a handler for it. The client sends this with the tag id when the X on a buff is clicked.
- Added a small buff API on `Player`: `AddBuff(iconId, nameId, durationMs, onEnded)` returns a tag id, `RemoveBuff` takes a buff off, and `CancelBuff` is what the X calls. Buffs expire on the player's update tick, and `onEnded` runs when a buff expires or is cancelled.
- Transformations from the boss cake, Scaredy Cake and transform foods show a buff with the item name, and cancelling it changes the player back.
- Food effects show a 30 minute buff and cancelling it removes the aura. The aura is now removed by the buff timer instead of a delayed packet.

Duration is in seconds and the icon id is an image id from Images.txt (`icon_status_changeform_32` is 3843). Buffs aren't sent again on login or saved yet, that can come when something longer lived like coin boosts needs it.


Sulphural:
image

Adding screenshot here.

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -723,34 +717,186 @@ public PlayerUpdatePacketAddPc GetAddPcPacket()
...
+
+        UnapplyEffect(effect);
+        effect.OnRemoved?.Invoke();
+    }
+
+    private void TickEffects(DateTimeOffset now)
+    {
+        List<PlayerEffect>? starting = null;
+        List<int>? expired = null;
+
```
yungcomputerchair, line 778: This function would be a lot easier to read if these lists were empty instead of null. I don't think you're really saving on perf by having these nullable

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -723,34 +717,186 @@ public PlayerUpdatePacketAddPc GetAddPcPacket()
...
+    }
 
-        SendTunneledToVisible(new PlayerUpdatePacketUpdateTemporaryAppearance { Guid = Guid, TemporaryAppearance = modelId }, true);
+    #region Effects
+
+    // A tracked, timed effect on the player - a transformation, a food effect aura, or both a
+    // world-visible composite effect and a buff bar icon at once. One tick loop expires them,
+    // one pair of methods applies/unapplies whatever packets each of those parts needs, so adding
+    // a new kind of effect doesn't mean adding another set of ad hoc fields and conditions.
+    private readonly Dictionary<int, PlayerEffect> _effects = new();
```
yungcomputerchair: Any reason you are using a normal `Dictionary` + a lock instead of a `ConcurrentDictionary`?

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -723,34 +717,186 @@ public PlayerUpdatePacketAddPc GetAddPcPacket()
...
+                SendTunneledToVisible(new PlayerUpdatePacketPlayCompositeEffect { Guid = Guid, CompositeEffectId = effect.AppearancePoofEffectId, Position = Position, Clear = false }, true);
+
+            SendTunneledToVisible(new PlayerUpdatePacketUpdateTemporaryAppearance { Guid = Guid, TemporaryAppearance = effect.AppearanceModelId }, true);
 
-        if (_temporaryAppearanceEffectId != 0)
+            // Unlike reverting to the full player model (see UnapplyEffect below), swapping into
+            // a transform model doesn't drop a world-visible effect tag that's already attached
+            // to the actor - it just keeps playing. Resyncing here anyway used to leave a second,
+            // orphaned copy of the tag behind once the original was cancelled, which is what kept
+            // the aura visible no matter what got removed.
```
yungcomputerchair: you can leave out the second part here
```suggestion
            // Unlike reverting to the full player model (see UnapplyEffect below), swapping into
            // a transform model doesn't drop a world-visible effect tag that's already attached
            // to the actor - it just keeps playing.
```

src/Sanctuary.Game/Entities/Player.cs
```diff
@@ -723,34 +717,186 @@ public PlayerUpdatePacketAddPc GetAddPcPacket()
...
+
+            SendTunneled(new ClientUpdatePacketAddEffectTag
+            {
+                TagId = effect.Id,
+                EffectTag = new EffectTag
+                {
+                    InstanceId = effect.Id,
+                    Duration = durationSeconds,
+                    StartTime = 0,
+                    StopTime = unchecked((uint)-durationSeconds)
```
yungcomputerchair: I'm a little concerned that `unchecked` is needed here; do you know why?

src/Sanctuary.Gateway/Helpers/Abilities/ConsumableAbility.cs
```diff
@@ -38,7 +36,7 @@ public abstract class ConsumableAbility(AbilityServices services)
 
     public abstract bool HandleAbility(Player player, AbilityPacketClientRequestStartAbility packet, int slot, ClientItem clientItem, ClientItemDefinition itemDefinition);
 
-    protected static int NextEffectTagId() => Interlocked.Increment(ref _castFxTagCounter);
+    protected static int NextEffectTagId() => EffectTagIdGenerator.Next();
```
yungcomputerchair, line 41: maybe leave a comment here about effect tag IDs being global and not scoped to the player

src/Sanctuary.Game/Entities/Player.cs
yungcomputerchair, line 87: is `_delayedPackets` still used?

yungcomputerchair (changes requested):
small suggestions and a couple questions, but this architecture looks loads better than before.

src/Sanctuary.Game/Entities/Player.cs
Sulphural: Turned out not needed. We doesn't set CheckForOverflowUnderflow, so the cast wraps the same way with or without unchecked. Dropped it.

src/Sanctuary.Game/Entities/Player.cs
Sulphural: Yes, still used. SillyStringAbility's reset animation and the boombox/cake poof delay in ConsumableAbility.PlayEffect both still go through it.

yungcomputerchair (approved):

PR #127, opened 2026-09-14
Reattach food effect aura when a temporary appearance reverts

Sulphural:
## Summary
- Ending a transformation (or a boombox dance) rebuilds the player actor client-side, which silently drops the food effect's composite effect tag. It now reattaches the aura once the model is back to normal instead of leaving it hidden until the next transformation.
- The client ignores an `AddEffectTagCompositeEffect` that reuses a tag id it has already seen for that actor, so the reapply mints a fresh id. Tag id generation moved into `Sanctuary.Game` so both the gateway abilities and `Player` share one counter.


yungcomputerchair:
The system design here is starting to feel a little finnicky to me. I feel like there are a lot of complex conditions in place in the code to account for a bunch of different scenarios, and whenever we add a new kind of effect we have to account for all the existing effects. The number of bugs we've had to fix already is indicative of the design slipping up somewhere, IMO.

Is there maybe a higher-level design that we can use to track these effects/transformations on the Player? Maybe a `Stack` of effects/transformations with timeouts that we can tick, calculate a composite state from, and whenever something changes we can broadcast update packets?

This would also allow us to strip out the delayed packet queue, too, since we'd just track the effects over time. Pretty similar to the buffs PR you have up right now alongside this.

Sulphural:
Closing this in favor of #126 - @yungcomputerchair's suggestion above (a generic effect stack instead of ad hoc per-effect fields) turned out to unify cleanly with the buff bar work, so I've folded this fix into that PR using the same underlying mechanism. #126 now supersedes this one.

PR #128, opened 2026-09-15
Add Ballet, Chiptune, Robot, Chicken, and Plants Dance Boomboxes

Sulphural:
## Summary
Five boomboxes existed in our data with models, dance animations, and (in some cases) icons already sitting in the client assets, but were never actually wired up to a real item or a store listing. This adds all five to the coin store:

- **Chicken Boombox** and **Plants Dance Boombox** turned out to already have real, orphaned marketplace listings elsewhere in `StoreBundles.json` (correct name, description, icon, and price), just never connected to an item definition or boombox behavior. Both are now fully correct, not placeholders.
- **Ballet**, **Chiptune**, and **Robot Boombox** don't have any real store text/price anywhere in this data or in public sources I could find (Ballet was staff-event-only, Chiptune only sold for ~5 weeks during 8-Bit Month 2013, Robot only dropped from a mystery chest). These three are functional (correct model, dance, price) but show placeholder name/description/icon in the store, clearly marked as such in the resource comments.

Also removed two pre-existing duplicate listings I found in the boombox section of the store while working on this (Realms Roll Boombox and Rainbow Randomizer Boombox were each listed twice).


cillow19 (approved):
if these new additions have been tested already, lgtm

PR #130, opened 2026-09-19
Add quest definitions, JSON loading and the CharacterQuests table

Sulphural:
Data model and resource loading for quests, plus the persistence schema they'll use. Quests are authored in Resources/Quests.json (text ids, giver/turn-in NPCs, rewards, ordered goals). QuestDefinitionCollection indexes quests by giver and goal target, and assigns world guids to Collect goal pickups.

Replaces #109, closed when this fork left the upstream network. Same branch, no changes.


src/Sanctuary.Game/Resources/Definitions/QuestDefinition.cs
```diff
@@ -0,0 +1,64 @@
...
+    // 0 = none.
+    public int PrerequisiteQuestId { get; set; }
+    public int NextQuestId { get; set; }
+
+    // Quests that block this one while active or completed, for mutually exclusive quests such as the
+    // two race-specific introductions. Each lists the other so the check is symmetric.
+    public List<int> ExcludesQuestIds { get; set; } = [];
+
+    // Badge on the giver: "!" while available, "?" while in progress.
+    public int NotificationAvailable { get; set; } = 2;
```
ImAlko, line 43: Are these two constants?

src/Sanctuary.Game/Resources/Definitions/QuestGoal.cs
```diff
@@ -0,0 +1,98 @@
...
+    public List<int> TargetResponseIds { get; set; } = [];
+
+    public int RequiredCount { get; set; }
+
+    // A CollectionNodeTypes.json key. Reuses the existing collection node system for pickups
+    // instead of quests owning their own spawn points.
+    public string CollectNodeType { get; set; } = string.Empty;
+
+    // [x, y, z]. The proximity check is 2D, so the Y only feeds the map pin.
+    public float[] ReachPosition { get; set; } = [];
```
ImAlko, line 52: Should this be a `Vector3` or whatever it is?

src/Sanctuary.Game/Resources/Definitions/QuestGoal.cs
```diff
@@ -0,0 +1,98 @@
...
+
+    // 0 = 12.
```
ImAlko, line 54: What is this comment implying?

src/Sanctuary.Game/Resources/Definitions/QuestDefinition.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    public int RewardCoins { get; set; }
+
+    // Goes to the player's active profile.
+    public int RewardExperience { get; set; }
+
+    // Item definition ids.
+    public List<int> RewardItems { get; set; } = [];
+
+    // 0 = none.
```
ImAlko, line 34: I am _honestly_ not sure what the better convention is,  so take this with a grain of salt, but shouldn't we just make this `int?` (i.e., nullable)?

src/Sanctuary.Game/Resources/Definitions/QuestDefinition.cs
```diff
@@ -0,0 +1,64 @@
...
+    // Shared by the offer flow and the giver's badge so the two can't drift.
+    public bool IsOfferableFor(IReadOnlyDictionary<int, bool> playerQuests)
+    {
+        if (playerQuests.ContainsKey(QuestId))
+            return false;
+
+        if (PrerequisiteQuestId != 0)
+            return playerQuests.TryGetValue(PrerequisiteQuestId, out var prerequisiteDone) && prerequisiteDone;
+
+        foreach (var excludedId in ExcludesQuestIds)
```
ImAlko, line 58: Don't we need to check excluded quests _before_ we check the pre-req? What happens if you have two mutually exclusive quest branches that share a pre-req and then you've completed one of them? The other one shouldn't be offered, but I _think_ (you'll need to double check me) that in your current case it _will_ be offered because the pre-req is complete.

src/Sanctuary.Game/Resources/Definitions/QuestGoal.cs
```diff
@@ -0,0 +1,98 @@
...
+    public int DescriptionId { get; set; }
+
+    // What the goal's NPC says on completion. Dialogue overrides this when set.
+    public int DialogueId { get; set; }
+    public List<QuestDialogueLine> Dialogue { get; set; } = [];
+
+    public QuestGoalType Type { get; set; } = QuestGoalType.TalkToNpc;
+
+    // 0 = the quest's TargetGuid.
+    public ulong TargetGuid { get; set; }
```
ImAlko, line 35: Something about having a `TargetGuid` and `TargetGuids` sits right with me. I think I can see _why_ you have it this way, but I'm not sure if its the right approach

src/Sanctuary.Game/Resources/Definitions/QuestGoal.cs
```diff
@@ -0,0 +1,98 @@
...
+
+    // Extra NPCs that also credit this goal, for a counted talk step where several interchangeable
+    // NPCs share one tracker row. RequiredCount sets how many are needed.
+    public List<ulong> TargetGuids { get; set; } = [];
+
+    // Per-NPC lines, aligned with AllTalkTargetGuids().
```
ImAlko, line 41: Isn't this fragile? Wouldn't just defining something like

```
public sealed class TalkTarget
{
    public ulong Guid { get; set; }
    public int DialogueId { get; set; }
    public int ResponseId { get; set; }
}
```

enforce things?

src/Sanctuary.Game/Resources/Definitions/QuestGoal.cs
```diff
@@ -0,0 +1,98 @@
...
+    // 0 = reuse NameId.
+    public int DescriptionId { get; set; }
+
+    // What the goal's NPC says on completion. Dialogue overrides this when set.
+    public int DialogueId { get; set; }
+    public List<QuestDialogueLine> Dialogue { get; set; } = [];
+
+    public QuestGoalType Type { get; set; } = QuestGoalType.TalkToNpc;
+
+    // 0 = the quest's TargetGuid.
```
ImAlko, line 34: What does this comment mean and if it is `0`, doesn't `AllTalkTargetGuids()` not run? Maybe that's okay, I just don't know what that the usecase is

ImAlko:
I'd like to suggest making `QuestGoal` abstract and having child classes like `TalkToNpcGoal`, `ReachLocationGoal`, `CollectGoal` that would split up the logic for the different types of goals. Are there other quest types that you'd be adding in the future? Something like complete <X> dungeon or complete <Y> minigame or something? That would make the `QuestGoal` class blow up with content, most of which will probably have irrelevant logic that is associated with all the other goals. 

I'm not sure if what I'm suggesting is correct/good but it's worth at least thinking about and getting input from other contributors on.

